PR #1, opened 2021-07-22
SCIENCE-1642: Develop Lambda job to trigger Glue workflow

philerooski:
https://sagebionetworks.jira.com/browse/SCIENCE-1642


philerooski:
The initial commit takes care of the meat of the implementation. There are still a few unknowns, such as how the SNS message will be formatted and whether there are any additional run properties which we would like to pass to the workflow.

philerooski:
At some point this was merged into main (with some changes)

PR #3, opened 2021-09-14
[ETL-33/55] Ready BridgeDownstream for infra development

tthyer:
This PR does some of the usual things we want to do when we're bootstrapping a repository for infrastructure development.

1. Creates a pipenv environment file (Pipfile) which will be used both in local development and for CICD.
2. Sets up sceptre structure and begins sceptre configuration for the deployment of AWS Cloudformation stacks.
3. Creates a "development bucket" template. For our development process, we'll use this bucket to stand in for Bridge's bucket where raw data will be dropped. This bucket is being deployed in both the develop and prod accounts; it doesn't belong in prod but it allowed me to put something in there so sceptre doesn't fail the deployment because sceptre's funny that way.
4. Creates a deployment workflow for CICD. This relies on Environment secrets defined in this github repo's settings and a matrix strategy which takes the environments into account. This is a new pattern at Sage and may be of interest to @BrunoGrandePhD and @zaro0508.

Also, a python-flavored .gitignore file has been added. I will add a CODEOWNERS file after discussing what are appropriate teams for IBC-DPE with @thomasyu888.


README.md
```diff
@@ -1 +1 @@
-# spark-mhealth-demo
\ No newline at end of file
+# spark-mhealth-demo
```
tthyer, line 1: Some files are getting touched by pre-commit  -- usually to edit white space or number of lines at end of file.

templates/development-bucket.yaml
```diff
@@ -0,0 +1,101 @@
...
+
+  BucketName:
+    Value: !Ref Bucket
+    Export:
+      Name: !Sub '${AWS::Region}-${AWS::StackName}-BucketName'
+
+  BucketArn:
+    Value: !GetAtt Bucket.Arn
+    Export:
+      Name: !Sub '${AWS::Region}-${AWS::StackName}-BucketArn'
```
tthyer, line 101: This is a _draft_ of the bucket we'll need. It may be complete, or not. We'll tweak this as we go -- it is a development bucket and does not need to be perfect now.

.github/workflows/aws-deploy.yaml
```diff
@@ -0,0 +1,70 @@
...
+      matrix:
+        job-environment:
+          - develop
+          - prod
+        include:
+          - job-environment: develop
+            sceptre-environment: develop
+          - job-environment: prod
+            sceptre-environment: prod
+    environment: ${{ matrix.job-environment }}
```
tthyer, line 29: Lines 19-29 here are a new pattern; I introduced this to reduce duplication when deploying to more than one environment (as we currently have in Sage-Bionetworks-Workflows/aws-workflows-nextflow-infra repo). This depends upon there being a "develop" and a "prod" Environment in the settings for this github repo, containing some environment-specific secrets.

I chose to keep "job-environment" and "sceptre-environment" separate for a couple reasons: first, they're actually different, that is github vs sceptre concepts, and so can diverge. If we conflate them now and someone decides the sceptre "develop" folder should really be called "dev", that will break the CICD. Second, I was testing out defining additional variables in matrix includes. That said I could definitely be convinced to use the following instead:
```
strategy:
  matrix:
    environment:
      - develop
      - prod
```

.github/workflows/aws-deploy.yaml
```diff
@@ -0,0 +1,70 @@
...
+on: push
+
+jobs:
+  pre-commit:
+    name: Run pre-commit hooks against all files
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/checkout@v2
+      - uses: actions/setup-python@v2
+      - uses: pre-commit/action@v2.0.2
```
tthyer, line 12: This would benefit form the creation of one of the new composite actions.

BrunoGrandePhD (approved):
I'm approving this new direction for our infra CI workflows. 🚀

philerooski (approved):

PR #8, opened 2021-09-20
[ETL-33] CFN stacks: revise bucket, add role, database, classifier

tthyer:
This is the next installment for ETL-33. 

1. There is a small reorganization. All python has been moved under the `src` directory, which does have subfolders underneath it. I moved diagrams to a docs directory.
2. The S3 bucket has been renamed for more generic use, and instead of one development bucket we have two: one each for input and output. In production we will only have an output bucket as will pull real data from a bucket in the Bridge account.
3. I've added a classifer that is probably complete, and a job role and glue database that both will  probably undergo some further changes


Pipfile
```diff
@@ -14,3 +14,4 @@ sceptre = "==2.*"
 boto3 = "==1.*"
 pyspark = "==3.*"
 synapseclient = "==2.*"
+pyyaml = "*"
```
tthyer, line 17: for linting

config/develop/glue-database.yaml
```diff
@@ -0,0 +1,6 @@
+template_path: glue-database.yaml
+stack_name: glue-database
+parameters:
+  DatabaseName: bridge-downstream-etl
+stack_tags:
+  {{ stack_group_config.default_stack_tags }}
```
tthyer, line 6: I anticipate that this will eventually not be a free-standing stack, but will rather be part of a nested stack created per-study.

templates/s3-bucket.yaml
```diff
@@ -76,7 +76,7 @@ Resources:
             - Sid: ReadAccess
               Effect: Allow
               Principal:
-                AWS: !Ref ReadWriteAccessArns
+                AWS: !Ref ReadOnlyAccessArns
```
tthyer, line 79: this was a mistake in the previous PR

templates/glue-database.yaml
```diff
@@ -0,0 +1,15 @@
+AWSTemplateFormatVersion: '2010-09-09'
+Description: TBD
```
thomasyu888, line 2: Nit: Fill in description later?

thomasyu888 (approved):
LGTM. just one small comment.

templates/glue-database.yaml
tthyer: thanks -- pushed a description

PR #9, opened 2021-09-21
Create ECR repository for use with lambda

philerooski:


tthyer (approved):
This looks like it should work fine. Ideally you will test-deploy this before merging, but this is simple enough that I don't think it's absolutely necessary.
You would do that by following [these directions](https://github.com/Sage-Bionetworks-Workflows/infra-template/blob/0b10c5a0c4d459339eb8e4232881c823e49bb89c/README.md#setup).

If you are creating a Docker image, we'll want to get that into CICD too. This may mean we have three separate workflows in our repository. I'm attempting to make that work, but for the sake of simplicity this is the sort of thing that will drive me to break apart the repositories.

philerooski:
@tthyer Will you please break apart the repositories as you see fit? I do not have a strong preference how we organize things.

I tried following the directions to test the stack and ran into a "No stacks detected from the given path..." error. I've given up for now! I have never worked with sceptre before.

PR #10, opened 2021-09-22
Etl-55 scripts workflow

tthyer:
This adds an additional workflow for the purpose of uploading python scripts to a bucket. This is necessary b/c the Glue jobs want to pull scripts from an S3 bucket.

This CICD workflow adds more complexity than I would like, but the current configuration of workflow scripts seems to be behaving itself for the moment.


thomasyu888:
👌

PR #11, opened 2021-09-23
Etl-55 scripts workflow part. 2

tthyer:
This is a refinement on the scripts workflow
1. Fixes an issue with the bucket directory setting
2. Alters the workflow to only trigger if path changes match pattern 'src/glue/jobs/*.py' rather than all py files. This will be necessary if we are going to keep the lambda in the same repository, or it will also trigger this workflow. It needs its own, separate workflow.


thomasyu888:
Lgtm

PR #12, opened 2021-09-24
Etl-66 Glue Jobs

tthyer:
1. This creates the set of jobs for a study.
2. This adds uploading templates as well as the scripts, purely so I can create a nested stack for glue jobs instead of having to repeat this a lot.
3. I unified the two CICD workflows because Github Actions has remarkably poor support for dependency between workflows, and we have to upload some files to buckets before we can use those in stacks, but excellent support for dependencies between jobs in a single workflow.

We had previously discussed these being shared; however, since the jobs which use json_s3_to_parquet.py require that a Database and Table be defined for use in the arguments and these are not shared, it appears to me that we will define a glue-jobs stack for each study. @philerooski, let me know if you think this is not the case. If we can dispense with the DefaultArguments in the jobs, or at least with `--database` and `--table`, then the jobs should be shareable (since we can also define properties for workflows).


config/develop/glue-jobs.yaml
```diff
@@ -0,0 +1,10 @@
+template_path: glue-jobs.yaml
+stack_name: glue-jobs
+parameters:
+  SourceBucketName: {{ stack_group_config.artifact_bucket_name }}
+  BranchOrTagName: ETL-66/glue-jobs
```
thomasyu888, line 5: Should this be switched to `main`

templates/glue-jobs.yaml
```diff
@@ -0,0 +1,146 @@
...
+    Properties:
+      TemplateURL: !Sub https://${SourceBucketName}.s3.amazonaws.com/${BranchOrTagName}/templates/glue-job.yaml
+      Parameters:
+        JobDescription: Export motion data in parquet format
+        GlueDatabaseName: !Ref GlueDatabaseName
+        GlueTableName: motion
+        JobName: MotionParquet
+        JobRole: !Ref JobRole
+        S3OutputBucketName: !Ref S3OutputBucketName
+        S3ScriptLocation: !Sub s3://${SourceBucketName}/${BranchOrTagName}/glue/jobs/json_s3_to_parquet.py
```
thomasyu888, line 122: So each of these jobs will create the different glue tables?

thomasyu888 (approved):
LGTM, minor comments

templates/glue-jobs.yaml
philerooski: No, the job names indicate which Glue table each job reads as input. They then relationalize the schema as defined in the glue table, producing multiple, flat tables, and writing each table to its own parquet dataset.

philerooski:
The jobs can be shared because the jobs are only parameterized by `table`. The all caps job parameters `WORKFLOW_NAME`, `WORKFLOW_RUN_ID`, and `JOB_NAME` are not passed by us, but are provided by Glue at runtime. All other parameters are _workflow_ parameters. Workflows are specific to each study, so this is where we define the database and the bucket output location.

> Are the correct jobs defined? When I pulled the jobs from sandbox, for example, there was both a json_s3_to_parquet job that uses the dataset_taskresult table, and a json_to_parquet_taskResult that uses the dataset_taskResult table. Which is correct?

The json_s3_to_parquet job was used during testing and is not part of the `json_to_parquet` workflow.

> Is the "weather" dataset valid, or was that a test?
weather is one of the json files included in each archive.

> Please verify values -- for examplle, most jobs use only two workers, but json_s3_to_parquet uses 10
json_s3_to_parquet isn't used in the workflow. Considering that we're likely dealing with less than 100 records each time we run the json_to_parquet workflow, and each job in that workflow deals with < ~25 MB worth of data, even two workers is overkill. We could try using just one worker per json_to_parquet job.

config/develop/glue-jobs.yaml
tthyer: yes thanks for catching that

templates/glue-jobs.yaml
tthyer: also, a tables PR will be forthcoming next

tthyer:
@philerooski thanks, here's what I'll do then:

1. I am going try removing the DefaultArguments of the jobs. These are in the job descriptions I pulled from sandbox, but based on your comments, I will see if I can remove them.
2. Keeping weather. Removing json_s3_to_parquet.
3. Reducing default workers from 2 to 1.

tthyer:
@philerooski Ok after looking at things and rereading your comments:
I still need to pass table names in the Job DefaultArguments: those are not being passed from anywhere else. 
However I think I understand what you mean: those table names won't change, just the db names will.

philerooski:
@tthyer Yes, that's exactly right.

PR #16, opened 2021-10-02
Add glue tables stack

tthyer:
This adds the tables stack; it probably needs to be updated. @philerooski Can you confirm which of the tables in sandbox are latest? Should I update from the tables in mtb_construct database?


thomasyu888 (approved):
When the schema changes, will these have to change too?

philerooski:
The mtb_construct database contains the most up to date schemas. I don't expect there are any changes from what is in phil_sample_db, though. The tables in mtb_construct are more comprehensive in the sense that I crawled files coming from multiple types of assessments (not just number match) to construct the schemas. But since schemas shouldn't be changing from assessment to assessment it makes no difference.

One important thing to keep in mind is that -- as part of the s3_to_json job/workflow -- all s3 object metadata is inserted into the metadata file. Hence all those fields should be represented in the metadata glue table schema. We don't yet have a clear picture of how that metadata will look. See https://sagebionetworks.jira.com/wiki/spaces/MTB/pages/2458681562/MTB+metadata+upload+for+Bridge+Exporter+3.0

PR #22, opened 2021-10-22
Etl-73 refactor

tthyer:
This refactors the infra for a change to triggers and crawlers where a new crawler and a new trigger are added. Acc. to @philerooski in ETL-73, "Glue requires partitions to be added to the glue table definition before they will be picked up as new partitions vis a vis the job bookmark. This is done by running a crawler over the relevant S3 locations."


templates/glue-crawlers.yaml
```diff
@@ -0,0 +1,102 @@
...
+Resources:
+
+  StandardCrawler:
+    Type: AWS::Glue::Crawler
+    Properties:
+      Configuration: '{"Version":1.0,"CrawlerOutput":{"Partitions":{"AddOrUpdateBehavior":"InheritFromTable"}},"Grouping":{"TableGroupingPolicy":"CombineCompatibleSchemas"}}'
+      DatabaseName: !Ref DatabaseName
+      Name: !Ref StandardCrawlerName
+      RecrawlPolicy:
+        RecrawlBehavior: CRAWL_NEW_FOLDERS_ONLY
```
thomasyu888, line 51: Will older folders need to ever be re-run?

PR #23, opened 2021-10-26
[ETL-71] Nested template for individual studies

tthyer:
This nested template makes use of many of the existing templates to create a set of resources dedicated to one specific study: specifically, crawlers, triggers, workflows, database, tables.

Two example study configs are created in develop to verify that parallel/separate sets of infrastructure are created for each, and that there's no conflict when deploying multiple instances of the template.


tthyer:
@philerooski I've made your recommendation about including the studyname in the path, but I'm stilll not sure whether what I've currently got is right. I'm going to leave this PR open until you and I can functionally test the pipeline next Tuesday.

PR #24, opened 2021-11-10
Etl-65/synapsify

tthyer:
This is the first of several PRs for [ETL-65](https://sagebionetworks.jira.com/browse/ETL-65) -- getting ready to convert the pipeline to being initiated from a lambda that listens for sns messages. 

This one takes the existing source bucket and connect it to a Synapse project and upload test data to it.
This requires modifying the bucket template, creating a script to do all the synapse bits.
Uses the existing CFN macro for S3objects.

Some problems:
- had to add pandas to my venv or generating the manifest file failed, even though pandas was in my global environment
- sync always uploads, unfortunately -- just going with that for a while since it's only a few MB. Will create a feature request for more fine-grained control.


tthyer:
@thomasyu888 @BrunoGrandePhD This is a kind of a long fiddly PR, most of which won't be interesting or make sense if you haven't been following the project, but the bit I'm most interested in getting feedback on is `setup_test_data.py` because it uses synapseformation.

I have an S3 bucket created through CFN. The script creates the project and connects bucket and project. It uses manifest and sync to Synapse. The script is invoked through an after_launch hook in the sceptre configuration.

src/scripts/setup_test_data/setup_test_data.py
```diff
@@ -0,0 +1,118 @@
...
+  # get synapse client
+  ssm_parameter = 'synapse-bridgedownstream-auth'
+  syn = get_synapse_client(ssm_parameter=ssm_parameter)
+
+  # see if project exists and get its id
+  principal_id = '3432808' # BridgeDownstream Synapse service account
+  project_id = get_project_id(syn, principal_id)
+  if not project_id:
+    template_path = './src/scripts/setup_test_data/synapse-formation.yaml'
+    project_id = create_project(syn, template_path)
```
thomasyu888, line 96: I suspect if you had to re-run this code, it would be fail because create_resource currently doesn't handle updating of resources.  An error is thrown if the project already exists.

thomasyu888 (commented):
LGTM (with one comment). I'll wait for Bruno to add his thoughts, but it's awesome to see these tools in use.  This is one of the ways we envisioned users to chain together these tools.

src/scripts/setup_test_data/setup_test_data.py
tthyer: If I had to make an *update*, yes, that would be a problem. I've been resorting to manually deleting the project in that case. As it is everything runs great b/c I check whether I was able to retrieve a project id and only run create_project if not.

So that leads me to several things I have been thinking I would like from synapseformation:
1. get_or_create instead of just create. I know of "only_get" but I think that requires that I edit the template or something...and I want the client to be smart for me, not to do this in my script
2. a warning for create when the project already exists instead of an error
3. an update method

thomasyu888: Ahh, yes I see that part of the code now. Thanks for the suggestions.

I think the client currently fills out the template with IDs and writes it back, I can see how that's not ideal as the client should just be smart enough to find the resources on synapse if they already exist (technically synapse can do this)

When you store entities with the same name in synapse, it will automatically get the entity if it already exists.  I should draft up some design docs for how synapseformation should work and have it as something the team can implement.

thomasyu888 (approved):

PR #25, opened 2021-11-17
Etl-65/lambda

tthyer:
This PR does two things:
1. Make the sns-to-glue lambda ready for production using AWS SAM.
2. Revise how the test data is set up so that S3 object metadata is set.


src/glue/jobs/s3_to_json_s3.py
```diff
@@ -14,7 +14,7 @@
 
 sc = SparkContext()
 glueContext = GlueContext(sc)
-log = glueContext.get_logger()
+logger = glueContext.get_logger()
```
tthyer, line 17: renaming for consistency's sake with other scripts

templates/glue-jobs.yaml
```diff
@@ -14,6 +14,13 @@ Parameters:
       this is used to build up the path to the correct object keys in the bucket
     Default: main
 
+  CodeRepositoryName:
+    Type: String
+    Description: >-
+      Name of code repository, used to build up the path to the correct
+      object keys in the bucket
+    Default: BridgeDownstream
+
```
tthyer, line 23: This is added because BridgeDownstream isn't the only thing writing to the template artifacts bucket.

src/lambda/template.yaml
```diff
@@ -0,0 +1,101 @@
...
+          Properties:
+            Topic: !Ref SnsTopic
+      Environment:
+        Variables:
+          SSM_PARAMETER_NAME:
+            Ref: SsmParameterName
+          GLUE_WORKFLOW_NAME:
+            Ref: WorkflowName
+      Timeout: 60
+      ImageUri: snstogluefunction:python3.8-v1
```
thomasyu888, line 76: Is this image auto generated?

src/scripts/setup_test_data/metadata.json
```diff
@@ -0,0 +1,30 @@
+{
+ "1Dbywpp4w2zhXjZPgItrKvlh-raw.zip" : {
+   "recordid": "1Dbywpp4w2zhXjZPgItrKvlh",
+   "substudymemberships": "|mtb-alpha=NNJ18004|",
```
thomasyu888, line 4: Out of curiosity, why the pipes?

src/scripts/setup_test_data/setup_test_data.py
```diff
@@ -26,13 +27,15 @@ def get_synapse_client(ssm_parameter):
     Name=ssm_parameter,
     WithDecryption=True)
   syn = synapseclient.Synapse()
-  syn.login(authToken=token['Parameter']['Value'])
+  synapse_auth_token = token['Parameter']['Value']
+  os.environ['SYNAPSE_AUTH_TOKEN'] = synapse_auth_token
+  syn.login(authToken=synapse_auth_token, silent=True)
```
thomasyu888, line 32: Is there a reason you do both?  Unless you call `syn.login` later on, I don't think you need to have both `authToken=...` and `os.environ....`

thomasyu888 (approved):

thomasyu888:
Awesome!  just some comments.

src/lambda/template.yaml
tthyer: `sam build` builds the images for the lambda and updates the template. Actually, this line is relic of when I had to go recover the template from `.aws_sam/build` dir -- I think I'll remove it to avoid confusion.

src/scripts/setup_test_data/metadata.json
tthyer: No clue. I was replicating test data. This is a Bridge thing and i don't think it's used at all.

src/scripts/setup_test_data/setup_test_data.py
tthyer: I think there was a version of synapseformation that required that I log in explicitly; I'm not recalling exactly now. I have to have a syn because I use that directly around the synapseformation call, and since create_synapse_resources takes a syn, I can just do away with the env var.

src/lambda/template.yaml
tthyer: And to be more clear, when the image is built it is this block that determines how the image gets labeled:
```
    Metadata:
      Dockerfile: Dockerfile
      DockerContext: ./sns_to_glue
      DockerTag: v0.1
```

src/scripts/setup_test_data/setup_test_data.py
tthyer: Wait, I'm just realizing that what I wrote doesn't make sense.

tthyer: I have the latest version of synapseclient (2.5.0) but setting the env var only didn't work for me, so I removed that.

PR #26, opened 2021-12-03
Etl-84/s3tojson tweaks

tthyer:
- What was the glue-job template becomes the glue-spark-job template, and there's a new glue-python-job template too: the first is for glueetl jobs (i.e. spark) and the 2nd is for pythonshell. These take different options, or more accurately, the latter takes a small subset of the former's possible options.
- s3_to_json_s3.py logging changed to python logging instead of spark logging
- Replaced old test data with updated data (more about this in the next PR that deals with ETL-83)
- Made the S3ToJson job able to handle multiple records (more groundwork for later) and had to extend its timeout to handle that


thomasyu888:
What are all the binary zip files?

thomasyu888:
Lgtm except for the failing deployment

philerooski (approved):

tthyer:
That failing deployment has happened a couple of times when github action runs `sam build`. Always fails with error `Error: SnsToGlueFunction failed to build: toomanyrequests: Rate exceeded` on the ECR pull. Going to try a docker login step to ECR to fix this b/c authenticated pulls have a diff rate limit than unauthenticated, and our github workflow must be using a shared IP, which would be the identity from ECR's p.o.v., so it's not too surprising that I would see this error periodically.

PR #27, opened 2021-12-06
Add dataset mapping and reference it in s3_to_json_s3 script

philerooski:
This addresses https://sagebionetworks.jira.com/browse/ETL-74

* Added a dataset mapping file. Right now the only app version present in the file is the most recent test build, version 1.0.2, build 57
* Added a function to get the dataset mapping file from S3 and reference it when bucketing the JSON file. The location of the dataset mapping file is assumed to be in the same bucket as the `s3_to_json_s3` job script (as deployed by sceptre) and the key is hardcoded to its deployment location. I updated the deployment configuration so that both glue jobs and their resources (just dataset_mapping file, so far...) are included in the sceptre deployment.

### DIFF
--- a/.github/workflows/upload-and-deploy.yaml
+++ b/.github/workflows/upload-and-deploy.yaml
@@ -55,8 +55,8 @@ jobs:
       - name: Copy files to templates bucket
         run: |
           # copy glue python scripts to bucket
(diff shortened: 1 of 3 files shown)


tthyer (approved):
lgtm

PR #29, opened 2021-12-09
ETL-85: unique job names / stack decoupling

tthyer:
Introduce a uniquish value for job names so that replacement can happen without dependent stacks complaining. This may be overkill since I'm also decoupling the stacks by replacing CFN ImportValue with sceptre params.


philerooski (approved):

tthyer:
@thomasyu888 fyi I'm going to merge this now to unblock @philerooski but would still like reviews from you guys

thomasyu888:
Lgtm

PR #33, opened 2021-12-17
Initial commit archive_dataset script

philerooski:
Addresses: https://sagebionetworks.jira.com/browse/ETL-89

Also added module comments to the top of the other scripts I wrote
in the past few weeks. Additionally, extended backfill_json_datasets
to use cached Synapse credentials if an ssm-parameter name is not
passed.

### DIFF
--- a/src/scripts/archive_dataset/archive_dataset.py
+++ b/src/scripts/archive_dataset/archive_dataset.py
@@ -1,0 +1,141 @@
+"""
+Archive a dataset and its derived datasets by copying them to the
+archive folder. A typical use case is to archive a dataset before a compatible
+schema change. Multiple compatible schema changes may happen to
+a single dataset, thus archives have an update number appended
+to their name. For example, if this is the third time we are
+archiving the `dataset_metadata_v2` dataset, the archive will be named
+`dataset_metadata_v2_3`. The script will automatically determine which
+update number this is, therefore it is not necessary to pass the
+update number as a command line argument.
+
+Datasets are assumed to exist in:
+
+    s3://{bucket}/{app}/{parquet}/
+
+And archives will be stored to:
+
+    s3://{bucket}/{app}/{parquet}/archive/
+
+When relationalizing a dataset, as is done in the json_to_parquet jobs,
+descendent datasets (such as `dataset_metadata_files`, a descendent
+of `dataset_metadata`) are created. For ease of use with the typical
+use case, descendent datasets are archived when passing the root
+dataset to `--dataset`. For example, passing `--dataset=dataset_metadata`
+and `--dataset-version=v2` will archive both `dataset_metadata_v2` and
+`dataset_metadata_files_v2`.
+
+Archive updates are tracked at the root dataset level.
+For example, if a struct `folders` is added to the `metadata_v2`
+JSON dataset in that dataset's second compatible schema update, then
+the next time a compatible schema update occurs the third archive
+operation will also occur. Although this is the first time that the
+`metadata_folders_v2` dataset will be archived, its archive will be
+named `metadata_folders_v2_3`.
+"""
+import argparse
+import shlex
+import subprocess
+import boto3
+
+
+def read_args():
+    parser = argparse.ArgumentParser(
+            description=("Archive a parquet dataset for each study of an app "
+                         "by copying it to the standard archive location: "
+                         "s3://{--bucket}/{--app}/{study}/parquet/archive/"
+                         "{--dataset}_{--dataset-version}_{update-number}/"))
+    parser.add_argument("--bucket",
+                        help="The name of the S3 bucket.")
+    parser.add_argument("--app",
+                        help="The app identifier.")
+    parser.add_argument("--dataset",
+                        help="The dataset name.")
+    parser.add_argument("--dataset-version",
+                        help="The dataset name.")
+    parser.add_argument("--profile",
+                        help="The AWS profile to use.")
+    args = parser.parse_args()
+    return args
+
+
+def get_source_and_dest_prefix(s3_client, bucket, app, dataset, dataset_version):
+    source_and_dest = dict()
+    study_prefix_obj = s3_client.list_objects_v2(
+            Bucket=bucket,
+            Prefix=f"{app}/",
+            Delimiter="/")
+    study_prefixes = [cp["Prefix"] for cp in study_prefix_obj["CommonPrefixes"]]
+    for study_prefix in study_prefixes:
+        parquet_prefix = f"{study_prefix}parquet/"
+        parquet_dataset_prefix_obj = s3_client.list_objects_v2(
+                Bucket=bucket,
+                Prefix=parquet_prefix,
+                Delimiter="/")
+        descendent_dataset_prefixes = [
+                cp["Prefix"] for cp in parquet_dataset_prefix_obj["CommonPrefixes"]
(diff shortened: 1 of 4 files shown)


src/scripts/archive_dataset/archive_dataset.py
```diff
@@ -0,0 +1,141 @@
+"""
+Archive a dataset and its derived datasets by copying them to the
+archive folder. A typical use case is to archive a dataset before a compatible
+schema change. Multiple compatible schema changes may happen to
+a single dataset, thus archives have an update number appended
+to their name. For example, if this is the third time we are
+archiving the `dataset_metadata_v2` dataset, the archive will be named
```
thomasyu888, line 7: Why is the version not increased instead when there is a schema change?

src/scripts/archive_dataset/archive_dataset.py
```diff
@@ -0,0 +1,141 @@
...
+                        help="The dataset name.")
+    parser.add_argument("--dataset-version",
+                        help="The dataset name.")
+    parser.add_argument("--profile",
+                        help="The AWS profile to use.")
+    args = parser.parse_args()
+    return args
+
+
+def get_source_and_dest_prefix(s3_client, bucket, app, dataset, dataset_version):
```
thomasyu888, line 62: Not a huge deal, but some semantics: Add function doc string, and look into python typing.  You can now add type hints to functions and I think it helps.

Also I can infer all the variables here because I read that large doc up top, but it would help to have variable descriptions in the doc string.

src/scripts/archive_dataset/archive_dataset.py
```diff
@@ -0,0 +1,141 @@
...
+                study_prefix=study_prefix,
+                dataset=dataset,
+                dataset_version=dataset_version)
+        for descendent_dataset_prefix in descendent_dataset_prefixes:
+            source_path = f"s3://{bucket}/{descendent_dataset_prefix}"
+            descendent_dataset = "_".join(
+                    descendent_dataset_prefix
+                    .split("/")[-2]
+                    .split("_")[:-1])
+            dest_path = (f"s3://{bucket}/{study_prefix}parquet/"
```
thomasyu888, line 90: Is there a typo here? Should it be `}/parquet`?  Also I recommend using `os.path.join`

src/scripts/archive_dataset/archive_dataset.py
```diff
@@ -0,0 +1,141 @@
...
+                         f"archive/{descendent_dataset}_{dataset_version}_"
+                         f"{latest_archive_dataset_update+1}/")
+            source_and_dest[source_path] = dest_path
+    return source_and_dest
+
+
+def get_archive_dataset_update_number(s3_client, bucket, study_prefix,
+                                      dataset, dataset_version):
+    archive_prefix = f"{study_prefix}parquet/archive/"
```
thomasyu888, line 99: Use os.path.join. it's risky to always assume someone that will pass in a study prefix that ends with a /

src/scripts/archive_dataset/archive_dataset.py
```diff
@@ -0,0 +1,141 @@
...
+            Delimiter="/")
+    relevant_archive_dataset_prefixes = []
+    if "CommonPrefixes" in archive_dataset_prefix_obj.keys():
+        relevant_archive_dataset_prefixes = [
+                cp["Prefix"] for cp in archive_dataset_prefix_obj["CommonPrefixes"]
+                if f"{dataset}_{dataset_version}" in cp["Prefix"]]
+    if len(relevant_archive_dataset_prefixes) == 0:
+        # No datasets of this type and version in archive
+        return 0
+    preexisting_update_nums = [
```
thomasyu888, line 112: This will break if the archive number doesn't get added on accident.

src/scripts/archive_dataset/archive_dataset.py
```diff
@@ -0,0 +1,141 @@
...
+            prefix.split("_")[-1][:-1] for prefix in relevant_archive_dataset_prefixes]
+    latest_version = max([int(n) for n in preexisting_update_nums])
+    return latest_version
+
+
+def copy_source_to_dest(source_and_dest):
+    for source in source_and_dest:
+        dest = source_and_dest[source]
+        bash_command = f"aws s3 cp --recursive {source} {dest}"
```
thomasyu888, line 121: Why not use `subprocess.check_call`?  Then you can use a list instead of f string interpolation

thomasyu888 (commented):
Some general comments for you. I like how you add examples in the docstrings - very informative. 

Not completely necessary to include the jira link in the description but I understand it does make it more convenient than searching for the issue after going to the ETL project in jira.

src/scripts/archive_dataset/archive_dataset.py
philerooski: Versions are increased in the case of an _incompatible_ schema change. This script is run during a compatible schema change, so no need to bump the version.

src/scripts/archive_dataset/archive_dataset.py
philerooski: No typo, since the prefixes are returned with a trailing `/`, but I will import the os module and join these so that the code can be more clear.

src/scripts/archive_dataset/archive_dataset.py
philerooski: `subprocess.run` was introduced in python3.5 to replace `subprocess.check_*` like methods. Both methods take a list.

thomasyu888: Ahh, I'm behind times a bit - thanks for letting me know!

src/scripts/archive_dataset/archive_dataset.py
philerooski: added function docstrings

src/scripts/archive_dataset/archive_dataset.py
```diff
@@ -0,0 +1,190 @@
...
+    Keyword arguments:
+    s3_client -- An AWS S3 client object.
+    bucket -- The S3 bucket name.
+    app -- The app identifier, as used in the S3 prefix in the bucket, to archive.
+    dataset -- The name of the dataset to archive.
+    dataset_version -- The dataset version to archive.
+
+    Returns:
+    A dictionary with source prefixes as keys and destination
+    prefixes as values.
```
thomasyu888, line 76: Just out of curiosity, where did you find this python docstring standard?  I'm only familiar with
* google
* pydoc
* numpy

src/scripts/archive_dataset/archive_dataset.py
```diff
@@ -0,0 +1,190 @@
...
+    source_and_dest -- A dictionary with S3 source prefixes as keys
+    and S3 destination prefixes as values.
+
+    Returns: None
+    """
+    for source in source_and_dest:
+        dest = source_and_dest[source]
+        bash_command = f"aws s3 cp --recursive {source} {dest}"
+        subprocess.run(
+                shlex.split(bash_command),
```
thomasyu888, line 172: This is very cool! I did not know about `shlex`

thomasyu888 (approved):
LGTM - just some comments.

src/scripts/archive_dataset/archive_dataset.py
philerooski: https://www.python.org/dev/peps/pep-0257/

tthyer (approved):

PR #35, opened 2022-01-19
Etl-83 Schema update changes

tthyer:
[ETL-83](https://sagebionetworks.jira.com/browse/ETL-83)

The primary purpose of this PR is to update the table schemas. A few fixes and additions were added to supporting scripts.

I will create a follow-up PR with practical remarks on the process of schema updates -- in particular how to go about debugging when there are issues.


src/scripts/setup_test_data/setup_test_data.py
```diff
@@ -98,6 +98,7 @@ def add_test_data(syn, dir_path, bucket_name, folder_id):
           with archive.open('metadata.json') as metadata_file:
             metadata = json.load(metadata_file)
             file_metadata['taskIdentifier'] = metadata['taskIdentifier']
+            file_metadata['appversion'] = metadata['appVersion']
```
tthyer, line 101: This script, which sets up the testing data, had to be updated for changes in S3ToJson Glue script.

philerooski (approved):

PR #38, opened 2022-01-21
Use uploadedOn rather than createdOn for derived year/month/day partition fields

philerooski:
Resolves ETL-97: https://sagebionetworks.jira.com/browse/ETL-97


thomasyu888 (approved):

tthyer (approved):
LGTM. We'll need to delete test data and recreate because metadata can only be added to S3 objects on creation.

PR #39, opened 2022-01-25
ETL-99/spark UI

tthyer:
This enables the Spark UI for Spark jobs -- the config options is a bit misnamed. Actually it just saves the spark logs so you can ingest them into Spark History Server. I also added CONTRIBUTING.md to explain how to spin up a docker container for that.

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -1,0 +1,10 @@
+This document is a work in progress.
+
+*Things to cover: pipenv, Cloudformation, Sceptre, SAM, Github workflows, git hooks*
+
+### Monitoring
+
+Use a combination of Workflows and Glue Studio, both within the [Glue console](https://console.aws.amazon.com/glue/), for graphical monitoring. Workflows gives a good view of what's happening within a workflow, but for a better overview of all the job runs, Glue Studio is recommended. Either option will give you a link to the Cloudwatch logs for each job.
+
+#### Debugging Spark Issues
+If your job is failing due to Spark issues, the Spark UI may give additional useful information. All jobs are configured to output spark logs. The current solution for viewing the Spark History Server is to run it locally through Docker. In order to do this, you need to be able to authenticate with AWS and get temporary credentials, and to run a docker image. Follow these [directions](https://docs.aws.amazon.com/glue/latest/dg/monitor-spark-ui-history.html#monitor-spark-ui-history-local) provided by AWS to start the server.
--- a/config/develop/glue-jobs.yaml
+++ b/config/develop/glue-jobs.yaml
@@ -6,7 +6,7 @@ dependencies:
 parameters:
   SourceBucketName: {{ stack_group_config.artifact_bucket_name }}
   JobRole: !stack_output_external glue-job-role::RoleArn
-  S3OutputBucketName: !stack_output_external bridge-downstream-dev-parquet-bucket::BucketName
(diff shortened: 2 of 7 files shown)


config/develop/glue-jobs.yaml
```diff
@@ -6,7 +6,7 @@ dependencies:
 parameters:
   SourceBucketName: {{ stack_group_config.artifact_bucket_name }}
   JobRole: !stack_output_external glue-job-role::RoleArn
-  S3OutputBucketName: !stack_output_external bridge-downstream-dev-parquet-bucket::BucketName
+  S3BucketName: !stack_output_external bridge-downstream-dev-intermediate-bucket::BucketName
```
tthyer, line 9: The name was changed because S3OutputBucketName is misleading -- this value is not actually what's being used to write the output. That's passed (confusingly) at the workflows level instead. But we pass in something here to write temp files too, as well as logs. Would be good eventually to be more consistent about this.

CONTRIBUTING.md
```diff
@@ -0,0 +1,10 @@
+This document is a work in progress.
+
+*Things to cover: pipenv, Cloudformation, Sceptre, SAM, Github workflows, git hooks*
+
+### Monitoring
+
+Use a combination of Workflows and Glue Studio, both within the [Glue console](https://console.aws.amazon.com/glue/), for graphical monitoring. Workflows gives a good view of what's happening within a workflow, but for a better overview of all the job runs, Glue Studio is recommended. Either option will give you a link to the Cloudwatch logs for each job.
```
tthyer, line 7: A fourth monitoring option will be the Cloudwatch metrics, but I'm still working on those.

templates/glue-spark-job.yaml
```diff
@@ -111,11 +111,15 @@ Resources:
         Name: !Ref JobCommand
         ScriptLocation: !Ref S3ScriptLocation
       DefaultArguments:
-        --TempDir: !Sub s3://${S3OutputBucketName}/tmp
+        --TempDir: !Sub s3://${S3BucketName}/tmp
         --enable-continuous-cloudwatch-log: !Ref ContinuousLog
+        --enable-metrics: true
+        --enable-spark-ui: true
+        --spark-event-logs-path: !Sub s3://${S3BucketName}/spark-logs/${AWS::StackName}/
```
tthyer, line 118: Lines 117 and 118 were the main reason for this PR

templates/glue-spark-job.yaml
```diff
@@ -111,11 +111,15 @@ Resources:
...
         --job-bookmark-option: !Ref BookmarkOption
         --job-language: !Ref JobLanguage
         --table: !Ref GlueTableName
+        # --conf spark.sql.adaptive.enabled
```
tthyer, line 122: This is something I am starting to tweak to control the number of partitions Spark is using, which determines the number of files output, unless you coalesce.

thomasyu888 (approved):

templates/glue-spark-job.yaml
philerooski: Are we writing the logs to the same bucket as our JSON datasets?

tthyer: Yes, to the "intermediate" bucket, which is basically the work bucket. Do you think that's problematic?

philerooski: No, that makes sense.

PR #43, opened 2022-02-09
Update sns_to_glue lambda to use Bridge SNS message format and SQS trigger

philerooski:
Resolves ETL-72 https://sagebionetworks.jira.com/browse/ETL-72

- The lambda handler was almost completely rewritten to receive messages from SQS rather than SNS.
- This required us to rewrite how we generate our test events.
- On the flip side, the lambda handler is now simpler, and we can probably stop building our own image for its dependencies.
- Synapse authentication was moved from the lambda handler to the s3_to_json_s3 Glue job.
- s3_to_json_s3 Glue job was modified to accept new workflow parameters as passed to it by the lambda handler.


config/develop/glue-job-role.yaml
```diff
@@ -1,4 +1,6 @@
 template_path: glue-job-role.yaml
```
zaro0508, line 1: the `template_path` key is deprecated, you should use `template` instead.  https://docs.sceptre-project.org/dev/docs/template_handlers.html#template-handlers

src/lambda/sns_to_glue/app.py
```diff
@@ -1,50 +1,37 @@
...
-def get_s3_loc(synapse_id, auth_token):
-    syn = synapseclient.Synapse()
-    syn.login(authToken=auth_token, silent=True)
-    f = syn.get(synapse_id, downloadFile=False)
-    logger.debug(f'synapse get response: {f}')
-    bucket = f['_file_handle']['bucketName']
-    key = f['_file_handle']['key']
-    s3_loc = {'bucket': bucket, 'key': key}
-    return s3_loc
+def lambda_handler(event, context):
```
zaro0508, line 5: ahh a lambda, i'm not sure if you are aware but Sceptre supports direct deployments of lambdas to AWS with the [sceptre-sam-handler](https://github.com/Sceptre/sceptre-sam-handler).  You may want to check it out as it may simplify/streamline your CI.

config/develop/glue-job-role.yaml
tthyer: @philerooski Let's create a separate JIRA ticket to make this update, but it's an easy one to do.

src/lambda/sns_to_glue/app.py
tthyer: @zaro0508 Thanks for telling us about this handler -- we'll make a JIRA to look into it. It may not be an easy switch for us since we are not using IT's standard process for uploading files to S3, and at first glance this handler only has a subset of the sam functionality. Have you updated the lambda-template repository to use it?

tthyer (approved):
LGTM! Thanks.

config/develop/glue-job-role.yaml
tthyer: created https://sagebionetworks.jira.com/browse/ETL-111

PR #45, opened 2022-02-14
Convert s3 to json s3 job from python shell to spark job

philerooski:
Resolves ETL-113: https://sagebionetworks.jira.com/browse/ETL-113

- S3ToJsonS3 becomes its own stack that can be shared among studies.
- Explicitly pass dataset mapping URI rather than relying on --scriptLocation argument
- Made some changes to the S3ToJsonS3 script so that it actually works (not sure how it was working before... failing silently maybe)


config/develop/glue-job-S3ToJsonS3.yaml
```diff
@@ -0,0 +1,16 @@
+template_path: glue-spark-job.j2
+stack_name: glue-job-S3ToJsonS3
+#{% set BranchOrTagName='main' %}
+parameters:
+  JobRole: !stack_output_external glue-job-role::RoleArn
+  S3BucketName: !stack_output_external bridge-downstream-dev-intermediate-bucket::BucketName
+  BookmarkOption: job-bookmark-disable
+  JobDescription: Convert data to JSONS3 data
+  MaxConcurrentRuns: '150'
+  S3ScriptLocation: s3://{{ stack_group_config.artifact_bucket_name }}/BridgeDownstream/{{ BranchOrTagName }}/glue/jobs/s3_to_json_s3.py
```
thomasyu888, line 10: Unsure, but can BranchOrTagName be used here?

config/develop/glue-job-S3ToJsonS3.yaml
```diff
@@ -0,0 +1,16 @@
+template_path: glue-spark-job.j2
+stack_name: glue-job-S3ToJsonS3
+#{% set BranchOrTagName='main' %}
```
thomasyu888, line 3: Does this job have dependencies?

templates/study-pipeline-infra.j2
```diff
@@ -107,33 +111,34 @@ Resources:
...
+        --enable-spark-ui: true
+        --spark-event-logs-path: !Sub s3://${JsonBucketName}/spark-logs/${AWS::StackName}/
+        --job-bookmark-option: job-bookmark-enable
+        --job-language: python
+        --table: {{ '{}'.format(dataset['table_name']) }}
+        # --conf spark.sql.adaptive.enabled
+      Description: {{ 'Export {} data in parquet format'.format(dataset['dataset_name']) }}
+      ExecutionProperty:
+        MaxConcurrentRuns: 1
+      GlueVersion: '3.0' # Spark 3.1.1, Python 3.7 # parameterize?
```
thomasyu888, line 133: Is the spark version linked to the Glue Version?

config/develop/glue-job-S3ToJsonS3.yaml
tthyer: No, not for this template. We're discussing a refactor for how this is handled to simplify it. Let's talk about it at the standup tomorrow.

config/develop/glue-job-S3ToJsonS3.yaml
tthyer: It does -- these are not called explicitly in this config but implicitly

PR #48, opened 2022-02-19
Use exporter 3.0 data and metadata

philerooski:
Resolves:
- [ETL-118](https://sagebionetworks.jira.com/browse/ETL-118) (which is duplicated by [ETL-116](https://sagebionetworks.jira.com/browse/ETL-116)) Create dataset mapping file which indexes dataset versions using new exporter 3.0 format
- [ETL-119](https://sagebionetworks.jira.com/browse/ETL-119) Update s3_to_json_s3 job to work with newly formatted dataset mapping
- [ETL-120](https://sagebionetworks.jira.com/browse/ETL-120) Update test data to use exporter 3.0 format

Summary:
- We are now using exporter 3.0 test data
- We map datasets to their versions by `osName` and `appVersion` which we parse from the `clientInfo` s3 metadata.
- We get test data metadata from a metadata manifest which is generated by a new script `src/scripts/curate_test_data/curate_test_data.py`, rather than reading metadata from within the test data archive.
- The `taskIdentifier` partition column became `assessmentid` because that's what Bridge is using. `taskIdentifier` is still included in the metadata dataset.
- Made a change to [.github/workflows/upload-and-deploy.yaml](https://github.com/Sage-Bionetworks/BridgeDownstream/compare/etl-119?expand=1#diff-1a4f4419dd4b3692946bb615fe4b24214acd0466ac4039bd2a126c5fc7f3f640) so that the build succeeds. We'll want to remove this once [aws-sam-cli](https://github.com/aws/aws-sam-cli/issues/3661) is patched.
- We used to hardcode an app version in the sceptre_user_data to pull specific datasets and their versions for sceptre to create resources from. Since the dataset mapping format changed I removed this from the config and updated the template which referenced this. It's still hardcoded (see 69baccc), but now we have more than one dataset mapping in the dataset mapping file, so eventually we'll have to come up with a more elegant solution.


tthyer (changes requested):

src/glue/jobs/s3_to_json_s3.py
```diff
@@ -72,27 +104,21 @@ def process_record(s3_obj, s3_obj_metadata, dataset_mapping):
...
                     j["day"] = int(uploaded_on.day)
                     for key in s3_obj_metadata:
-                        # We revert partition fields back to camelCase
-                        if key == "taskidentifier":
-                            j["taskIdentifier"] = s3_obj_metadata[key]
-                        elif key == "recordid":
-                            j["recordId"] = s3_obj_metadata[key]
-                        else:
-                            j[key] = s3_obj_metadata[key]
+                        j[key] = s3_obj_metadata[key]
```
thomasyu888, line 107: what is `j` here?

src/scripts/curate_test_data/curate_test_data.py
```diff
@@ -0,0 +1,65 @@
...
+    return metadata
+
+def main():
+    args = read_args()
+    syn = synapseclient.login()
+    table = syn.tableQuery(
+            args.query.format(source_table=args.file_view)).asDataFrame()
+    new_test_data = table.id.apply(syn.get).values
+    create_metadata_manifest(table=table)
+    if args.overwrite:
```
thomasyu888, line 54: You're assuming here that the path at which you execute the `create_metadata_manifest` is not the `data_dir`.  You can accidentally delete `metadata.json` that you wrote.

src/scripts/curate_test_data/curate_test_data.py
```diff
@@ -0,0 +1,65 @@
...
+        for f in os.listdir(args.data_dir):
+            os.remove(os.path.join(args.data_dir, f))
+    for file in new_test_data:
```
thomasyu888, line 57: Not necessary: might be more clear if you renamed `file` to `file_ent`.

src/scripts/setup_test_data/metadata.json
```diff
@@ -1,30 +1 @@
...
-   "recordid": "R7mBYI63NpCo3sJa9l0qnpza",
-   "substudymemberships": "|mtb-alpha=NBS22002|",
-   "healthcode": "nzaLFo5nhlqHc1TyUKhvsBP7",
-   "createdon": "1625238358060",
-   "taskidentifier": "MTB Spelling Form 1"
- },
- "Le9-WBxplHhGXo_FLNiY3SaI-raw.zip": {
-   "recordid": "Le9-WBxplHhGXo_FLNiY3SaI",
-   "substudymemberships": "|mtb-alpha=SBM50004|",
-   "healthcode": "XrvTSTKobPJAbxGwZ4sceNBf",
```
thomasyu888, line 26: I see that the metadata.json is pretty different now.  Do we expect this file to change over time?  How does this file affect the code?

src/scripts/setup_test_data/setup_test_data.py
```diff
@@ -107,33 +107,32 @@ def add_test_data(syn, dir_path, project_id):
...
-          name=filename,
-          synapseStore=False,
-          dataFileHandleId=file_handle['id'])
-        syn.store(file)
+      record_id = filename.split('-raw.zip')[0]
+      with open(f'{dir_path}/metadata.json') as metadata_file:
+        metadata = json.load(metadata_file)
+        for k, v in metadata[record_id].items():
+            if v is not None:
+                file_metadata[k] = str(v)
```
thomasyu888, line 115: Why is it important that everything is a string?

src/glue/jobs/s3_to_json_s3.py
```diff
@@ -51,11 +52,42 @@ def get_dataset_mapping(dataset_mapping_uri):
...
+    else:
+        app_version = app_version_search.group().split("=")[1]
+    if os_name_search is None:
+        print(client_info_str)
+    else:
+        os_name = os_name_search.group().split("=")[1]
+    client_info = {
+            "appVersion": app_version,
+            "osName": os_name}
+    return client_info
```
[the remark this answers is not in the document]
philerooski: This is why: https://sagebionetworks.jira.com/browse/BRIDGE-3222

src/scripts/curate_test_data/curate_test_data.py
```diff
@@ -0,0 +1,65 @@
+'''
+This script accepts a synapse table query and overwrites the contents of
+the test data folder with the records matched by the query. The table being
+queried is assumed to be a file view scoped over a "Raw Bridge Data" folder
+which has all annotations included in its schema. A metadata manifest is
+written in JSON format to the working directory. This is meant to be used
+with setup_test_data.py.
```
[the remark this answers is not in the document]
philerooski: At the top of this script is a global variable which lists all the fields I drop from the file view before writing the metadata. This is all Synapse file entity metadata and tells us exactly where the data came from. Would writing these fields and the query to a separate metadata_provenance.json file cover your concerns?

src/glue/jobs/s3_to_json_s3.py
philerooski: It's a JSON file in the data archive.

src/scripts/curate_test_data/curate_test_data.py
philerooski: The solution here is of the pick your poison variety. Moving the `create_metadata_manifest` call below the conditional would preserve the manifest.json in the case someone were to run the script from the data directory, but now the metadata is treated as regular data by the `setup_test_data` script and that's going to break the pipeline. Probably the cleanest solution is to explicitly pass the metadata output path as a command line argument.

src/scripts/curate_test_data/curate_test_data.py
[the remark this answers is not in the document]
tthyer: Yes, if metadata_provenance.json is appended to for each separate test dataset or there is a separate provenance file for each test dataset such that the entire suite of test data can be recreated easily. Would that be possibly with the information you would write to this file? If not, what would the level of effort be?

src/scripts/setup_test_data/metadata.json
philerooski: This file is used by `setup_test_data`, which adds every metadata field to the metadata file and only the partition fields to the rest of the data. @tthyer If we remove recordId from the partition fields this will change so that recordId is still added to every file, it's just not used as a partition field. We may want to consider adding other useful fields to every file. For example, science team is used to having participant identifiers and data groups included with every record. The way the system is designed now requires joins on the metadata to include this information. Something to discuss with science...

philerooski: >Do we expect this file to change over time?

Not really, but it's changing now because Bridge is slowly incorporating all the metadata they planned to include with exporter 3.0 data.

src/scripts/setup_test_data/setup_test_data.py
philerooski: S3 metadata has to be a string.

src/scripts/curate_test_data/curate_test_data.py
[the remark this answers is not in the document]
philerooski: Yes, those fields include Synapse IDs and versions which uniquely identifies a piece of data on Synapse.

src/glue/jobs/s3_to_json_s3.py
```diff
@@ -51,11 +52,42 @@ def get_dataset_mapping(dataset_mapping_uri):
...
 
 def process_record(s3_obj, s3_obj_metadata, dataset_mapping):
     uploaded_on = datetime.strptime(s3_obj_metadata["uploadedon"], '%Y-%m-%dT%H:%M:%S.%fZ')
+    client_info = parse_client_info(s3_obj_metadata["clientinfo"])
```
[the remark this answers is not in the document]
philerooski: Changed the function name to remove the ambiguity

src/scripts/curate_test_data/curate_test_data.py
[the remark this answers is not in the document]
philerooski: I'm now writing a data provenance file which is formatted the same as the metadata file except has all the non-metadata fields + recordId.

src/scripts/curate_test_data/curate_test_data.py
philerooski: I added a command line argument --metadat-dir which specifies where to write the metadata and data provenance files.

tthyer (approved):
LGTM

PR #49, opened 2022-02-24
Etl-121

tthyer:
This PR fulfills the needs of [ETL-121](https://sagebionetworks.jira.com/browse/ETL-121) and [ETL-122](https://sagebionetworks.jira.com/browse/ETL-122) but goes beyond those:

**Introduces the concept of "namespace" within an environment.** The same stack can be deployed multiple times in an environment if you use a different namespace than the default value and the stack makes uses of namespace to make itself unique (for example to set unique name for itself and its child resources through namespace). This will necessarily be more involved for more stacks than others. In some cases, it's very simple, it just affects one name. In other cases, just as the study stack, it affects everything down to the scripts in the glue workflow, including the paths they write to in S3.

**Introduces the use of [composite github actions]**(https://docs.github.com/en/actions/creating-actions/creating-a-composite-action). I've created one that does [some common setup](https://github.com/Sage-Bionetworks/action-pipenv-aws-setup) I do over and over when building infra. Doubtless we have others that we can make, such as our three pre-commit steps. 

**Actually automates test deployments with branch pushes.** Previously, our sceptre deployments to develop were only occurring manually. Now, with each branch push, a deployment is made to develop. If stacks are namespaced, when pushes are made the namespace will be set to the branch name. 

For example: let's say I create a namespaced stack called `{{ namespace }}-glue-foojob`. The default name of this job is going to be `bridge-downstream-glue-foojob`. But while I'm developing it, I created it on branch ETL-121, so the CICD will create stack `ETL-121-glue-foojob`.

**A new workflow has been introduced called "cleanup"** which happens on delete events. This will be triggered on branch deletion. Branches will be deleted automatically because in this repository pull requests have been set up to automatically delete head branches. There is still the flaw in this plan that if you create a branch, push it, and just leave it hanging around forever, there will be unnecessary stacks created in development that are undesirable, so we should discuss that.

Other details:
1. Workflows stack was unnested (so `glue-workflow.yaml` template is deleted). We no longer have any nested workflows -- all resources are created directly in `study-pipeline-infra.j2`.
2. I have a `manage_artifacts/artifacts.py` script that does what it says -- the upload/deletion of scripts and templates that our stacks depend upon. The github workflows reference this in some steps.


.github/workflows/upload-and-deploy.yaml
```diff
@@ -24,45 +24,36 @@ jobs:
     name: Upload files to S3 bucket
     runs-on: ubuntu-latest
     needs: pre-commit
-    if: startsWith(github.ref, 'refs/pull') != true
+    if: github.event_name == 'push' #startsWith(github.ref, 'refs/pull') != true
```
tthyer, line 27: I was playing with this. I am playing with the idea of what needs to happen at all on pull request events -- I think everything important is on either push or delete.

.github/workflows/upload-and-deploy.yaml
```diff
@@ -24,45 +24,36 @@ jobs:
...
-        uses: aws-actions/configure-aws-credentials@v1
-        with:
-          aws-access-key-id: ${{ secrets.CI_USER_ACCESS_KEY_ID }}
-          aws-secret-access-key: ${{ secrets.CI_USER_SECRET_ACCESS_KEY }}
-          aws-region: us-east-1
-          role-to-assume: ${{ secrets.CI_ROLE_TO_ASSUME }}
-          role-duration-seconds: 1200
+      - name: Set namespace for main
+        if: github.ref_name == 'main'
+        run: echo "NAMESPACE=bridge-downstream" >> $GITHUB_ENV
```
philerooski, line 45: Why create this exception for the main branch? Seems like it would be more intuitive to simply use "main" as the namespace.

src/glue/jobs/s3_to_json_s3.py
```diff
@@ -125,6 +125,7 @@ def process_record(s3_obj, s3_obj_metadata, dataset_mapping):
                 with open(output_path, "w") as f_out:
                     json.dump(j, f_out, indent=None)
                     s3_output_key = os.path.join(
+                        workflow_run_properties["namespace"],
```
philerooski, line 128: Why do we add this to the workflow run properties? It's already in the workflow name and we don't actually do anything with it in the job.

philerooski (approved):

thomasyu888:
Can you comment on why we need to support multiple separate development stacks.

.github/workflows/upload-and-deploy.yaml
thomasyu888: Agreed - although we can discussed that this means edits to readmes would trigger this step as well.

.github/workflows/upload-and-deploy.yaml
thomasyu888: This leads me to think about the generalization of this.  If we have NAMESPACES as the github branch, if we had other projects that utilized this paradigm, would we have competing cloudformation stack names?

src/scripts/manage_artifacts/artifacts.py
```diff
@@ -0,0 +1,67 @@
...
+  print(f'Invoking command: {cmd}')
+  os.system(cmd)
+
+
+def delete(namespace):
+  '''Removes all files recursively for namespace'''
+  s3_path = f's3://{cfn_bucket}/{repo_name}/{namespace}/'
+  cmd = f'aws s3 rm --recursive {s3_path}'
+  print(f'Invoking command: {cmd}')
+  os.system(cmd)
```
thomasyu888, line 45: Nit: recommend using `subprocess.run` over `os.system`.  More description [here](https://docs.python.org/3/library/os.html#os.system)

src/scripts/manage_artifacts/artifacts.py
```diff
@@ -0,0 +1,67 @@
...
+def list_namespaces():
+  '''List all namespaces'''
+  s3_path = f's3://{cfn_bucket}/{repo_name}/'
+  cmd = f'aws s3 ls {s3_path}'
+  print(f'Invoking command: {cmd}')
+  os.system(cmd)
+
+
+def main(namespace):
+  namespace = args.namespace
```
BrunoGrandePhD, line 57: Where is `args` coming from? Should the function parameter be `args`?

BrunoGrandePhD (commented):
🚀

.github/workflows/upload-and-deploy.yaml
tthyer: The namespace is used to name the stack. I simply didn't like my stacks named "main-glue-foojob".  "bridge-downstream-glue-foojob" is more meaningful and official sounding to me. "ETL-121-glue-foojob" sounds to me like a stack that is in development. That was my reasoning.

`NAMESPACE` is being deliberately overloaded: it is being used as a stack prefix, and it is being used to construct artifact paths for S3. I deliberately overloaded it so we wouldn't have to pass two different things, both an acceptable`stack-prefix` variable and a `namespace` variable. We might even run the danger that those turn into four different things.

If you are adding a stack and you are merging it to main and it should be a unique stack, then it will get a unique name.

src/glue/jobs/s3_to_json_s3.py
tthyer: Can you be more clear with what you want to happen here instead?

tthyer:
> LGTM, but can you comment on why we need to support multiple separate development stacks.

@thomasyu888 In general, when developing infrastructure one needs to support multiple development stacks because otherwise each of you is attempting to deploy the same stack into the same environment, overwriting each other's work, and causing endless confusion. We are sort of providing a way to branch infrastructure.

.github/workflows/upload-and-deploy.yaml
tthyer: @thomasyu888 I am going to file a JIRA for these workflows to only trigger if the config or src directories are edited

src/glue/jobs/s3_to_json_s3.py
philerooski: I originally misunderstood the purpose of this line. It seems fine.

philerooski (approved):

src/scripts/manage_artifacts/artifacts.py
tthyer: you're right, and I think that's more than a nit :)

.github/workflows/upload-and-deploy.yaml
```diff
@@ -75,55 +66,64 @@ jobs:
...
+      - name: Setup code, pipenv, aws
+        uses: Sage-Bionetworks/action-pipenv-aws-setup@v1
+        with:
+          aws_access_key_id: ${{ secrets.CI_USER_ACCESS_KEY_ID }}
+          aws_secret_access_key: ${{ secrets.CI_USER_SECRET_ACCESS_KEY }}
+          role_to_assume: ${{ secrets.CI_ROLE_TO_ASSUME }}
 
-  sceptre-deploy:
+      - name: Create directory for remote sceptre templates
+        run: mkdir -p templates/remote/
```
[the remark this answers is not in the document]
tthyer: thanks -- I have a JIRA ticket to make that switch, and I'll link your example there

src/scripts/manage_artifacts/artifacts.py
tthyer: that was a woops -- thanks for pointing it out

PR #56, opened 2022-03-04
Use the correct ref name from the delete event when cleaning up

tthyer:
The problem with the cleanup.yaml has been that the `ref_name` field of the DELETE event's payload is _not_ the name of the branch being deleted -- it is `main`. To get the name of the branch being deleted, use `event.ref`.


thomasyu888 (commented):
Lgtm

PR #58, opened 2022-03-21
Etl-91 Job / artifact versioning

tthyer:
This lets us version jobs / artifacts. Note, this takes those out of the namespacing.
Removes "UniqueId" -- these aren't needed after all.
The CICD is rewritten  to be responsive to tagging. One tag has already been created for the purpose of testing, and we'll create a new tag after merging this PR and as needed. No auto-tagging has been set up.


.github/workflows/upload-and-deploy.yaml
```diff
@@ -40,12 +40,12 @@ jobs:
...
           version: 1.37.0
 
-      - name: Set namespace for branch
+      - name: Set namespace for non-default branch or for tag
         if: github.ref_name != 'main'
         run: echo "NAMESPACE=$GITHUB_REF_NAME" >> $GITHUB_ENV
 
       - name: Copy files to templates bucket
-        run: python src/scripts/manage_artifacts/artifacts.py --upload --namespace ${{ env.NAMESPACE }}
+        run: python src/scripts/manage_artifacts/artifacts.py --upload --ref $GITHUB_REF_NAME
```
tthyer, line 48: So namespaces are being applied for templates but refs are being used for artifacts/paths

config/develop/namespaced/glue-job-S3ToJsonS3.yaml
```diff
@@ -1,4 +1,5 @@
 #{% set ns = stack_group_config.namespace %}
+#{% set glue_script_tag = 'ETL-91' %}
```
tthyer, line 2: The glue_script_tag above we will change to an acceptable tag before it reaches prod. If the scripts were in a separate repository they would go through review, be merged to main, then tagged. Then in the infra repository, we would refer to the tagged version of the scripts. In this repo since we're trying to do everything in one place, the workflow will look something like:
1. do work on a branch for dev
2. merge to main
3. tag
4. fix up prod config including referencing correct tag for scripts
5. merge to main again

Eventually, we'll split all of this into separate repositories for simplicity and cleanliness.

philerooski (approved):

PR #60, opened 2022-03-24
[ETL-107] Add SQS template and config

philerooski:
This is just the SQS CFN template and an example config. The other work we've been doing (including moving the lambda to its own stack) will be included in the PR for ETL-117.

### DIFF
--- a/config/develop/sqs-queue.yaml
+++ b/config/develop/sqs-queue.yaml
@@ -1,7 +1,0 @@
-template_path: sqs-queue.yaml
-stack_name: sqsqueue
-parameters:
-  PrimaryQueueName: my_queue
-  DeadLetterQueueName: my_dead_letter_queue
-stack_tags:
-  {{ stack_group_config.default_stack_tags }}
--- a/config/develop/sqs-queue.yaml
+++ b/config/develop/sqs-queue.yaml
@@ -1,0 +1,7 @@
+template_path: sqs-queue.yaml
+stack_name: sqsqueue
+parameters:
+  PrimaryQueueName: my_queue
+  DeadLetterQueueName: my_dead_letter_queue
+stack_tags:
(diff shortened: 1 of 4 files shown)


config/develop/sqs-queue.yaml
```diff
@@ -0,0 +1,7 @@
+template_path: sqs-queue.yaml
+stack_name: sqsqueue
+parameters:
+  PrimaryQueueName: my_queue
```
tthyer, line 4: Let's give this a more descriptive name

config/develop/sqs-queue.yaml
```diff
@@ -0,0 +1,7 @@
...
+  DeadLetterQueueName: my_dead_letter_queue
```
tthyer, line 5: and another more descriptive name

tthyer (changes requested):
Looks good, I just need a few more descriptive names.

What did you do with the lambda-specific commits to the etl-107 branch that we were making?

philerooski:
@tthyer I namespaced the SQS queue stack and borrowed the existing naming convention we've used for namespaced resources. I also added some SNS resources and policies to the SQS template in anticipation of removing those same resources from the lambda template.

templates/sqs-queue.yaml
```diff
@@ -4,24 +4,14 @@ Description: >
...
-
 Resources:
 
   PrimaryQueue:
     Type: AWS::SQS::Queue
     Properties:
       DelaySeconds: 0
       MessageRetentionPeriod: 86400
-      QueueName: !Ref PrimaryQueueName
+      QueueName: !Sub '${AWS::StackName}-Queue'
```
tthyer, line 14: better

tthyer (approved):

PR #61, opened 2022-03-30
[ETL-117] Lambda polls from SQS queue

philerooski:
* Lambda has been pulled out of the study stack and is now in its own stack
* "sns to glue" nomenclature has been changed to "sqs to glue"
* Both Lambda and SQS queue are namespaced
* Lambda has permissions to invoke any glue workflow, not just one associated with a specific study or namespace
* The commit which used to be in this branch which changed the test events from SQS events to SNS events has been omitted (i.e., they're SQS events again). I'm not sure why I thought we would be passing those test events to the SQS queue rather than our lambda...

### DIFF
--- a/config/config.yaml
+++ b/config/config.yaml
@@ -1,5 +1,6 @@
 project_code: bridge_downstream
 namespace: {{ var.namespace | default('bridge-downstream') }}
+latest_version: v0.1
 region: us-east-1
 artifact_bucket_name: sceptre-cloudformation-bucket-bucket-65ci2qog5w6l
 synapseAuthSsmParameterName: synapse-bridgedownstream-auth
@@ -9,6 +10,6 @@ default_stack_tags:
   Project: mobile-toolbox
   OwnerEmail: aws-mobilehealth-dataengineering-dev@sagebase.org
 j2_environment:
-   extensions:
-      - jinja2.ext.do
-      - jinja2.ext.debug
+  extensions:
+    - jinja2.ext.do
+    - jinja2.ext.debug
--- a/src/lambda/template.yaml
+++ b/src/lambda/template.yaml
@@ -66,7 +66,7 @@ Resources:
             BatchSize: 20
             MaximumBatchingWindowInSeconds: 20
             Queue: !Ref SQSQueueArn
-      Timeout: 600
+      Timeout: 30
       Environment:
         Variables:
           NAMESPACE: !Ref Namespace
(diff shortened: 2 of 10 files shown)


templates/sqs-queue.yaml
```diff
@@ -30,18 +30,17 @@ Resources:
...
             AWS: '*'
           Action:
           - SQS:SendMessage
-          Resource: !Ref PrimaryQueue
+          Resource: !GetAtt PrimaryQueue.Arn
           Condition:
-            ArnEquals:
-              'aws:SourceArn': !Ref SnsTopic
-
+            ArnLike:
```
philerooski, line 42: `ArnLike` is how AWS does it when you use the web UI to connect the SQS queue as an event source for Lambda. I couldn't get `ArnEquals` to work for some reason. But according to the docs `ArnLike` and `ArnEquals` behave identically. So.... ¯\_(ツ)_/¯

templates/sqs-queue.yaml
```diff
@@ -81,3 +80,10 @@ Resources:
       Protocol: sqs
       TopicArn: !Ref SnsTopic
       Endpoint: !GetAtt PrimaryQueue.Arn
+
+Outputs:
+
+  PrimaryQueueArn:
```
philerooski, line 86: We pass this output to the lambda stack.

src/lambda/template.yaml
```diff
@@ -41,19 +41,32 @@ Resources:
             - glue:StartWorkflowRun
             - glue:PutWorkflowRunProperties
             Resource:
-            - !Sub arn:aws:glue:${AWS::Region}:${AWS::AccountId}:workflow/${WorkflowName}
+            - !Sub arn:aws:glue:${AWS::Region}:${AWS::AccountId}:workflow/*
```
thomasyu888, line 44: Why the switch to all workflows here?

thomasyu888 (approved):
Small comment. LGTM though

config/config.yaml
```diff
@@ -1,5 +1,6 @@
 project_code: bridge_downstream
 namespace: {{ var.namespace | default('bridge-downstream') }}
+latest_version: v0.1
```
thomasyu888, line 3: I like that you did this

src/lambda/template.yaml
philerooski: It used to be the case that we had one lambda per study. We now have one lambda per app. So a lambda will poll an SQS queue which is subscribed to the app-wide Bridge SNS topic, receiving messages for data from all different types of studies. The lambda then groups these messages by study and sends them to the appropriate study-specific workflow.

PR #63, opened 2022-04-08
[ETL-170] Generate test events from Synapse Dataset

philerooski:
See ticket for changes made: https://sagebionetworks.jira.com/browse/ETL-170

TLDR: We no longer store test data in this repo nor in an S3 bucket owned by us. Instead we reference Synapse datasets containing Bridge raw data entities to generate the test events.


src/scripts/setup_test_data/setup_test_data.py
```diff
@@ -50,112 +80,133 @@ def get_project_id(syn, principal_id):
...
-      with open(file_path, 'rb') as f:
-        bucket.put_object(
-          Body=f,
-          Key=obj_key,
-          Metadata=file_metadata
+    logger.error(e)
+    sys.exit(1)
+
+
+class TempFileView():
```
thomasyu888, line 95: ooo, I like this a lot!

src/lambda/events/generate_test_event.py
```diff
@@ -42,81 +45,111 @@
     "UnsubscribeURL": "https://sns.us-east-1.amazonaws.com/?Action=Unsubscribe&SubscriptionArn=arn:aws:sns:us-east-1:563295687221:phil_s3_bucket_update:89b889dd-9e68-41a5-b865-c194f09d024c"
 }
 
-def create_message_template(project_id, folder_id):
+def create_message_template(entity):
   return {
     "appId": "example-app-1",
-    "recordId": "-4I2GOqDSdjaXsbuw8oYXBKK",
+    "recordId": entity.annotations.recordId.pop(),
```
thomasyu888, line 51: Are there multiple record ids? or why are you doing a pop() here?

src/lambda/events/generate_test_event.py
```diff
@@ -42,81 +45,111 @@
...
     "record": {
-      "parentProjectId": project_id,
-      "rawFolderId": folder_id,
-      "fileEntityId": "",
-      "s3Bucket": "",
-      "s3Key": ""
+      "parentProjectId": "syn0000000",
```
thomasyu888, line 53: Does it matter this is syn000000? I think you can do entity.projectId?

src/lambda/events/generate_test_event.py
```diff
@@ -42,81 +45,111 @@
...
+
+def get_dataset_id_by_name(syn, parent_id, dataset_name):
+  response = list(syn.getChildren(parent_id, includeTypes=['dataset']))
+  dataset = next(item for item in response if item['name'] == dataset_name)
+  dataset_id = '' if dataset is None else dataset.get('id')
+  if not dataset_id:
+    print(f'No dataset {dataset_name} exists')
     sys.exit(1)
+  return dataset_id
+  return dataset
```
thomasyu888, line 105: Two return statements?

src/lambda/events/generate_test_event.py
```diff
@@ -42,81 +45,111 @@
...
+
+
+def main():
+  args = read_args()
+  syn = get_synapse_client(ssm_parameter=SSM_PARAMETER)
+  test_dataset_id = get_dataset_id_by_name(
+          syn=syn,
+          parent_id=args.synapse_project_id,
+          dataset_name=args.dataset_name)
+  test_dataset_metadata = syn.restGET(f"/entity/{test_dataset_id}")
```
thomasyu888, line 131: Nit: Can you add a comment here saying Datasets not supported which is why its a `restGET`

thomasyu888 (approved):
Some comments, but overall looks good.

src/lambda/events/generate_test_event.py
philerooski: For some reason every annotation is a list of one item. I don't know if Bridge is doing something peculiar to make the annotations formatted that way or something else is happening.

thomasyu888: Ya.. i'm not sure why this is the case either, but a list is what is returned. I was just curious if there was going to ever be more than one recordId.

src/lambda/events/generate_test_event.py
philerooski: `projectId` is a default view (dataset) column, but I pass the entity to this function so I don't have access to that information. If we actually used `parentProjectId` we could add a parameter to this function and/or iterate through the dataset rows rather than the entities. But that makes the preceding logic (line 135 `test_data = ...`) a little messier.

src/lambda/events/generate_test_event.py
philerooski: Whoops. I'll lint my changes and commit.

src/lambda/events/generate_test_event.py
philerooski: Sure.

src/lambda/events/generate_test_event.py
philerooski: No, that should be impossible.

thomasyu888:
@philerooski Feel free to merge after you fix the two return statements

PR #65, opened 2022-04-22
Changes necessary for exporting study data in prod

philerooski:
Change summary:
* Move artifact bucket from global config to dev/prod specific config
* Introduce prod stacks
* Add `microphone_v1` table. (Same structure as `microphone_levels_v1`, the developers decided to rename this file for some reason).
* Added Android build 14 and iPhone OS build 74 to dataset mapping
* Add bootstrap_trigger script (formerly backfill json datasets)
* Add crontab for bootstrap trigger script to `src/ec2/resources/`
* Added `JsonToParquetTriggerSchedule` parameter to study-pipeline template. Omitting this parameter creates an on-demand json to parquet trigger. Passing a [cron expression](https://docs.aws.amazon.com/glue/latest/dg/monitor-data-warehouse-schedule.html) to this parameter creates a json to parquet trigger which runs on a schedule. 6. t4tg.  6edrfgvb


config/prod/config.yaml
```diff
@@ -0,0 +1 @@
+artifact_bucket_name: sceptre-cloudformation-bucket-bucket-10mwvvuhlvtk9
```
thomasyu888, line 1: Two different buckets between develop and prod?

config/prod/studies/mobile-toolbox-htshxm.yaml
```diff
@@ -0,0 +1,30 @@
...
+  StudyName: htshxm
+  TemplateBucketName: {{ stack_group_config.artifact_bucket_name }}
+  ArtifactRef: {{ stack_group_config.latest_version }}
+  JsonBucketName: !stack_output_external bridge-downstream-intermediate-bucket::BucketName
+  ParquetBucketName: !stack_output_external bridge-downstream-parquet-bucket::BucketName
+  RoleArn: !stack_output_external glue-job-role::RoleArn
+  ClassifierName: !stack_output_external '{{ stack_group_config.namespace }}-array-of-records-classifier::ClassifierName'
+  SynapseAuthSsmParameterName: '{{ stack_group_config.synapseAuthSsmParameterName }}'
+  S3ToJsonS3JobName: !stack_output_external '{{ stack_group_config.namespace }}-glue-job-S3ToJsonS3::JobName'
+  JsonToParquetTriggerSchedule: 'cron(35 * * * * *)'
```
thomasyu888, line 21: Why different schedules?

src/scripts/bootstrap_trigger/Dockerfile
```diff
@@ -0,0 +1,4 @@
+FROM python:3.9.10
+
+RUN pip install boto3==1.16.33 pandas==1.1.5 synapseclient==2.5.1 pyarrow==4.0.0
+RUN git clone -b etl-129 https://github.com/Sage-Bionetworks/BridgeDownstream.git /root/BridgeDownstream
```
thomasyu888, line 4: will update this branch after it gets merged?

src/scripts/bootstrap_trigger/bootstrap_trigger.py
```diff
@@ -0,0 +1,176 @@
...
+
+
+def read_args():
+    parser = argparse.ArgumentParser(
+            description=("Submit files on Synapse to an AWS --glue-workflow"))
+    parser.add_argument("--file-view",
+                        required=True,
+                        help=("The Synapse ID of a file view containing "
+                              "files to be submitted."))
+    parser.add_argument("--raw-folder-id",
```
thomasyu888, line 25: What is the folder structure?  How do you have to deal with nested folder structure?

templates/ec2-bootstrap-trigger.yaml
```diff
@@ -0,0 +1,113 @@
+AWSTemplateFormatVersion: '2010-09-09'
+
+Description: >-
```
thomasyu888, line 3: I see, so theres an ec2 triggering the AWS glue jobs?

templates/ec2-bootstrap-trigger.yaml
```diff
@@ -0,0 +1,113 @@
...
+          aws ecr --region us-east-1 get-login-password | docker login --username AWS --password-stdin 634761300905.dkr.ecr.us-east-1.amazonaws.com/bootstrap_trigger
+          docker pull 634761300905.dkr.ecr.us-east-1.amazonaws.com/bootstrap_trigger
+          aws s3 cp ${CrontabURI} /tmp/crontab
+          crontab -u ec2-user /tmp/crontab
+
+  SecurityGroup:
+    Type: AWS::EC2::SecurityGroup
+    Properties:
+      GroupDescription: !Sub "The security group for ${AWS::StackName}"
+      SecurityGroupIngress:
```
thomasyu888, line 52: Why this security group?  also there probably needs to be tighter control on this instance.  We should probably put it behind our VPN.

thomasyu888 (changes requested):
Mostly looks good, but there is a security concern about the ec2.  We need to put it behind our VPN.

templates/ec2-bootstrap-trigger.yaml
thomasyu888: Adding VpcId - something like this: https://github.com/nlpsandbox/nlpsandbox-infra/blob/main/sceptre/nlpsandbox/templates/nlpsandbox.yaml#L188..

Do we have a VPC set up in this?

config/prod/config.yaml
philerooski: Yes.

config/prod/studies/mobile-toolbox-htshxm.yaml
philerooski: I'm concerned about running into the S3 write API limits. It's happened before where we were writing too many files to S3 at once and the job broke.

templates/ec2-bootstrap-trigger.yaml
philerooski: I made another commit to add the security group to the VPC

templates/ec2-bootstrap-trigger.yaml
philerooski: Yes.

src/scripts/bootstrap_trigger/bootstrap_trigger.py
philerooski: The folder structure doesn't matter. We only use this value to get an STS token. The file entity interfacing happens on a file view.

src/scripts/bootstrap_trigger/Dockerfile
philerooski: I added a commit which removes this branch reference.

config/prod/ec2-bootstrap-trigger.yaml
```diff
@@ -0,0 +1,7 @@
+template_path: ec2-bootstrap-trigger.yaml
+stack_name: ec2-bootstrap-trigger
+parameters:
+  SsmParameterName: synapse-bridgedownstream-auth
+  PrivateKey: ec2-bootstrap-trigger
+  CrontabURI: s3://{{ stack_group_config.artifact_bucket_name }}/BridgeDownstream/{{ stack_group_config.latest_version }}/ec2/resources/crontab
+  VPCId: vpc-d6c2a9ab
```
thomasyu888, line 7: What is this `vpc-d6c2a9ab`?  Did you create it yourself?

philerooski: You can view the VPC in the console. I didn't create it myself.

thomasyu888: I took a look at this, and the VPC is just the default VPC created and it doesn't meet our security standards.  This resource needs to sit behind the Sage VPN.

templates/ec2-bootstrap-trigger.yaml
philerooski: @thomasyu888 Does the EC2 need to be in a VPC because of the ingress rules? I should be able to remove this rule so that it's not possible to connect to the instance. Everything is taken care of by the EC2 user data. The ssh access is just for debugging purposes.

templates/ec2-bootstrap-trigger.yaml
```diff
@@ -116,3 +116,5 @@ Resources:
                 - ec2.amazonaws.com
             Action:
               - 'sts:AssumeRole'
+      ManagedPolicyArns:
+      - arn:aws:iam::aws:policy/AmazonSSMManagedInstanceCore
```
thomasyu888, line 120: ETL-175

config/prod/ec2-bootstrap-trigger.yaml
thomasyu888: See this: https://github.com/nlpsandbox/nlpsandbox-infra/blob/2155e67a552cb7cdaf9237e0bf889ac3f901ea3c/sceptre/nlpsandbox/templates/nlpsandbox.yaml#L121-L127

templates/ec2-bootstrap-trigger.yaml
```diff
@@ -32,13 +28,10 @@ Resources:
       InstanceType: t3.medium
       ImageId: ami-03ededff12e34e59e
       IamInstanceProfile: !Ref IAMInstanceProfile
-      KeyName: !Ref PrivateKey
-      SecurityGroupIds:
-        - !Ref SecurityGroup
+      SubnetId: !Ref SubnetId
```
thomasyu888, line 31: Follow example here: https://github.com/nlpsandbox/nlpsandbox-infra/blob/2155e67a552cb7cdaf9237e0bf889ac3f901ea3c/sceptre/nlpsandbox/templates/nlpsandbox.yaml#L121-L127

src/scripts/bootstrap_trigger/Dockerfile
```diff
@@ -0,0 +1,4 @@
+FROM python:3.9.10
+
+RUN pip install boto3==1.16.33 pandas==1.1.5 synapseclient==2.5.1 pyarrow==4.0.0
+RUN git clone -b https://github.com/Sage-Bionetworks/BridgeDownstream.git /root/BridgeDownstream
```
thomasyu888, line 4: Make sure the docker build doesn't build from an existing cache or new changes to the repo won't get pulled in.

templates/ec2-bootstrap-trigger.yaml
thomasyu888: @philerooski the only thing left is to not only assign this instance a private subnet but to follow the example above to assign it a security group and VPC.

philerooski: The subnet belongs to the VPC, so putting the EC2 in the subnet puts it in the VPC as well. I see there are multiple types of security groups associated with the VPC. The instance ends up in the default security group for the VPC, but I'm not sure which is the right security group to use.
image

thomasyu888: You'll want the VpnSecurityGroup.

thomasyu888 (approved):

PR #66, opened 2022-04-26
[ETL-166] De-dockerize lambda

philerooski:
* Lambda is built as a .zip file archive.
* Resources related to the lambda docker image were removed


thomasyu888 (approved):
LGTM

PR #68, opened 2022-04-28
[ETL-177] Update lambda README and copy relevant info to setup_test_data.py

philerooski:
Once @mattfazza is able to try their hand at invoking the lambda with the test events we can incorporate their feedback into this PR before merging.

### DIFF
--- a/src/lambda/README.md
+++ b/src/lambda/README.md
@@ -19,29 +19,46 @@ You may need the following for local testing.
 
 ## Use the SAM CLI to build and test locally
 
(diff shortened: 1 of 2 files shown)


thomasyu888:
I'm asking to be made maintainer of the BridgeDownstream team so that I can add Matt to the team giving them access.

philerooski:
@thomasyu888 Tess is the only maintainer of that team so it looks like you're out of luck. I did grant Matt write access directly which I think is sufficient.

thomasyu888:
Matt reviewed and tested this ticket.

PR #74, opened 2022-05-16
[ETL-198] Fix parquet diff and simplify bootstrap trigger

philerooski:
* Fix and simplify parquet diff process
* Improved argument descriptions and defaults

### DIFF
--- a/src/scripts/bootstrap_trigger/bootstrap_trigger.py
+++ b/src/scripts/bootstrap_trigger/bootstrap_trigger.py
@@ -18,6 +18,9 @@ from pyarrow import fs, parquet
 def read_args():
     parser = argparse.ArgumentParser(
             description=("Submit files on Synapse to an AWS --glue-workflow"))
+    parser.add_argument("--glue-workflow",
+                        required=True,
+                        help="The name of the Glue workflow to submit to.")
     parser.add_argument("--file-view",
                         required=True,
                         help=("The Synapse ID of a file view containing "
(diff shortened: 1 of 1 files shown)


src/scripts/bootstrap_trigger/bootstrap_trigger.py
```diff
@@ -28,30 +31,29 @@ def read_args():
...
+                        help=("Optional. The field name in the --file-view to diff "
+                              "upon. Defaults to 'recordId'."),
+                        default="recordId")
     parser.add_argument("--profile",
-                        help="The AWS profile to use.")
+                        help=("Optional. The AWS profile to use. Uses the default "
+                              "profile if not specified."))
     parser.add_argument("--ssm-parameter",
-                        help=("The name of the SSM parameter containing "
+                        help=("Optional. The name of the SSM parameter containing "
```
thomasyu888, line 56: Nit: https://stackoverflow.com/questions/24180527/argparse-required-arguments-listed-under-optional-arguments.  This isn't necessary, but just some food for thought - Positional arguments are usually "required" arguments vs the argparse default "optional arguments" when you use the `-h` flag.

src/scripts/bootstrap_trigger/bootstrap_trigger.py
```diff
@@ -71,17 +73,18 @@ def get_synapse_client(ssm_parameter=None, aws_session=None):
...
-def get_synapse_df(syn, entity_view, query=None):
+def get_synapse_df(syn, entity_view, index_field, query=None):
     if query is not None:
         query_string = query.format(source_table=entity_view)
     else:
         query_string = f"select * from {entity_view}"
     synapse_q = syn.tableQuery(query_string)
     synapse_df = synapse_q.asDataFrame()
-    return synapse_df
+    synapse_df_with_index = synapse_df.set_index(index_field)
```
thomasyu888, line 83: Ah, you're setting the index with a particular column name. +1

thomasyu888 (approved):

src/scripts/bootstrap_trigger/bootstrap_trigger.py
philerooski: tbh I have to disagree with argparse's reasoning. Why would they list flagged arguments with required=True under "optional arguments" just because there is a convention that flagged arguments be optional? I prefer flagged arguments to positional since they make explicit what is being set. With 3+ required positional arguments things start to get confusing.

thomasyu888: I agree with you.  I feel that if there is a solution that is provided via that stackoverflow, that they could easily create that argparse group for "required" named parameters, instead of having users create their own group.

PR #75, opened 2022-05-16
[ETL-199] Activate JSON to Paquet scheduled trigger upon creation

philerooski:


thomasyu888 (approved):

thomasyu888:
Interesting - these schedules don't start by default?

philerooski:
They're not "activated" by default. I guess that's so you can deploy scheduled triggers before you need them and then activate them once certain preconditions are met. In our case we simply want them to run right away.

PR #81, opened 2022-06-26
ETL-182

mattfazza:

### DIFF
--- a/.github/workflows/upload-and-deploy.yaml
+++ b/.github/workflows/upload-and-deploy.yaml
@@ -92,7 +92,7 @@ jobs:
           version: 1.37.0
 
       - name: sam build lambda
-        run: sam build --use-container
+        run: sam build
 
       - name: Invoke Lambda
         run:  sam local invoke -e events/single-record.json --env-vars test-env-vars.json
--- a/.github/workflows/upload-and-deploy.yaml
+++ b/.github/workflows/upload-and-deploy.yaml
@@ -92,7 +92,7 @@ jobs:
           version: 1.37.0
 
       - name: sam build lambda
-        run: sam build
+        run: sam build -s src/lambda/ -t src/lambda/template.yaml
 
       - name: Invoke Lambda
         run:  sam local invoke -e events/single-record.json --env-vars test-env-vars.json
--- a/.github/workflows/upload-and-deploy.yaml
+++ b/.github/workflows/upload-and-deploy.yaml
@@ -73,7 +73,7 @@ jobs:
     runs-on: ubuntu-latest
     needs: [pre-commit, upload-files, sceptre-deploy-branch]
     env:
-      NAMESPACE: bridge-downstream #$GITHUB_REF_NAME after testing hard-coded
+      NAMESPACE: $GITHUB_REF_NAME after
       LEMBDA_ENV_VARS_FILE: namespace_vars.json
     if: github.ref_name != 'main' && github.ref_type != 'tag'
     environment: develop
--- a/.github/workflows/upload-and-deploy.yaml
+++ b/.github/workflows/upload-and-deploy.yaml
@@ -73,7 +73,7 @@ jobs:
     runs-on: ubuntu-latest
     needs: [pre-commit, upload-files, sceptre-deploy-branch]
     env:
-      NAMESPACE: $GITHUB_REF_NAME after
+      NAMESPACE: $GITHUB_REF_NAME
       LEMBDA_ENV_VARS_FILE: namespace_vars.json
     if: github.ref_name != 'main' && github.ref_type != 'tag'
     environment: develop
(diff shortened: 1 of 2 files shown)


src/scripts/validate_workflow_run/test_validate_workflow_run.py
```diff
@@ -0,0 +1,45 @@
...
+import boto3
+from validate_workflow_run import read_args, get_json_by_key, get_parquet_by_key
+import pyarrow
+
+bad_bucket = 'xbridge-downstream-dev-intermediate-data'
+bad_key = 'x'
+good_json_bucket = 'bridge-downstream-dev-intermediate-data'
+good_json_key = 'ETL-91/example-app-1/study-1/raw_json/dataset=info_v1/assessmentid=dccs/year=2022/month=2/day=15/recordid=dsXJVIY0aFE5rAUdI3vpsGsr/dsXJVIY0aFE5rAUdI3vpsGsr.ndjson'
+good_parquet_bucket = 'bridge-downstream-dev-parquet'
+good_parquet_key = 'bridge-downstream/example-app-1/study-1/parquet/dataset_info_v1_files/assessmentid=ChangeLocalizationV1/year=2022/month=3/day=7/part-00154-7be7f212-ad59-4816-81b9-110401edfb4e.c000.snappy.parquet'
```
thomasyu888, line 11: Nit:  all caps global variables.

thomasyu888:
Couple of questions, but looks good so far

* What is capfd?
* Is there a reason why the tests shouldn't live in a "test" dir?

.github/workflows/upload-and-deploy.yaml
```diff
@@ -68,6 +68,35 @@ jobs:
...
+    runs-on: ubuntu-latest
+    needs: [pre-commit, upload-files, sceptre-deploy-branch]
+    env:
+      NAMESPACE: $GITHUB_REF_NAME
+      LEMBDA_ENV_VARS_FILE: namespace_vars.json
+    if: github.ref_name != 'main' && github.ref_type != 'tag'
+    environment: develop
+    steps:
+
+      - name: Setup code, pipenv, aws
```
philerooski, line 82: These first two steps are duplicated on line 29 and 36. Let's move them to their own job and include it in depends.

.github/workflows/upload-and-deploy.yaml
```diff
@@ -68,6 +68,35 @@ jobs:
...
+          aws_secret_access_key: ${{ secrets.CI_USER_SECRET_ACCESS_KEY }}
+          role_to_assume: ${{ secrets.CI_ROLE_TO_ASSUME }}
+
+      - name: Setup sam
+        uses: aws-actions/setup-sam@v1
+        with:
+          version: 1.37.0
+
+      - name: sam build lambda
+        run: sam build -s src/lambda/ -t src/lambda/template.yaml
```
philerooski, line 95: I wonder if we are going to need to change this if we ever have more than one lambda.

philerooski (approved):
One change suggested. Can you also run your python scripts through a linter? I use pylint, but you can use whatever your want.

.github/workflows/upload-and-deploy.yaml
thomasyu888: Will that work?  I feel like the environment of each gh-action step doesn't propogate.  Maybe I'm misunderstanding here.

src/scripts/validate_workflow_run/test_validate_workflow_run.py
mattfazza: Good catch, I'll change that.

.github/workflows/upload-and-deploy.yaml
mattfazza: Tom and I were under the same impression here.  I do like your approach to cache the dependencies across runs.  I'll add the cache: pipenv to it.

.github/workflows/upload-and-deploy.yaml
mattfazza: If we have more than one lambda and include it in the same template, this should still work.  It depends on how we want to implement additional lambdas - if we want to do that.

.github/workflows/upload-and-deploy.yaml
thomasyu888: Thanks both, I filed a JIRA: https://sagebionetworks.jira.com/browse/IBCDPE-290

thomasyu888 (approved):

thomasyu888:
@philerooski can we merge this?

philerooski:
@thomasyu888 I'm using this column types in ETL-238. I'm still testing whether I can avoid the issues in ETL-238 by changing the column type (for example, from a float to a double). So I might modify this branch later.

PR #83, opened 2022-07-08
[ETL-232] Update setup_test_data script

philerooski:
add one record of each assessment revision for each build to test dataset


tthyer (approved):
LGTM.

PR #85, opened 2022-07-20
[ETL-237] Update table_columns.yaml to better reflect JSON Schemas

philerooski:


thomasyu888 (approved):

thomasyu888:
Do we need to mechanism to make sure that these table columns better reflect our schemas in the future?  or is this a one time thing side from the potential schema changes coming down the pipe?

philerooski:
> Do we need to mechanism to make sure that these table columns better reflect our schemas in the future? or is this a one time thing side from the potential schema changes coming down the pipe?

Building Glue table schemas from JSON Schema is a complicated procedure. Tess spent some time trying to convert JSON Schema to Avro since Glue already supports instantiating Glue Schemas from Avro. But it required too much engineering effort.

PR #94, opened 2022-10-06
[ETL-23] Unit tests

philerooski:
* Unit tests for both S3 to JSON and JSON to Parquet Glue jobs
* Docker image used for testing environment is built and pushed to ECR every time.
* Glue job scripts refactored so that things aren't exposed to the global namespace (everything is in a function now)
* Test data added to `tests/test_json_s3_to_parquet` and `tests/data`.
* Dockerfile for testing environment added to `tests`.


tests/test_s3_to_json_s3.py
```diff
@@ -0,0 +1,477 @@
...
+        assert isinstance(schema_mapping, dict)
+
+    def test_get_archive(self, artifact_bucket, namespace):
+        archive_map = get_archive_map(archive_map_version="v4.4.1")
+        assert isinstance(archive_map, dict)
+
+    def test_update_sts_tokens(self):
+        synapse_data_folder_1 = "syn11111111"
+        sts_tokens = update_sts_tokens(
+                syn=MockSynapse(),
```
thomasyu888, line 293: This is one way to do it.  But you can actually use the `patch.Mock()` class for stuff like this.  or `patch.mock`.

philerooski: Nice! I didn't know about mock.

thomasyu888:
Can you add a testing section in the readme to explain why you tested things with Docker instead of installing things into Github Actions?

tests/test_s3_to_json_s3.py
```diff
@@ -0,0 +1,477 @@
...
+            "deviceInfo",
+            "deviceTypeIdentifier",
+            "files"
+          ]
+        }
+        return metadata_json_schema
+
+    def test_get_dataset_mapping(self, dataset_mapping_uri):
+        data_mapping = get_data_mapping(data_mapping_uri=dataset_mapping_uri)
+        assert isinstance(data_mapping, dict)
```
thomasyu888, line 280: Nit: Think about if you wanted to assert equals to a dictionary.

tests/test_s3_to_json_s3.py
```diff
@@ -0,0 +1,477 @@
...
+        assert cached_json_schema == "schema"
+
+    def test_get_json_schema(self, archive_map):
+        file_metadata_no_schema = {
+                "assessment_id": "spelling",
+                "assessment_revision": 5,
+                "file_name": "jellybeanz.json",
+                "app_id": "mobile-toolbox"
+        }
+        json_schema_no_schema = get_json_schema(
```
thomasyu888, line 314: One thing to be more explicit is to do:

```
s3_to_json_s3.get_json_schema
```
to distinguish between the functions/fixtures you have within your test.

tests/test_s3_to_json_s3.py
```diff
@@ -0,0 +1,477 @@
...
+    def test_get_cached_json_schema(self):
+        json_schema = {"url": "url", "schema": "schema", "app_id": None, "assessment_id": None,
+                "assessment_revision": None, "file_name": None, "archive_map_version": None}
+        cached_json_schema = _get_cached_json_schema(
+            url="url",
+            json_schemas=[json_schema]
+        )
+        assert cached_json_schema == "schema"
+
+    def test_get_json_schema(self, archive_map):
```
thomasyu888, line 307: Look into pytest parametrize

tests/test_s3_to_json_s3.py
```diff
@@ -0,0 +1,477 @@
...
+            assert file_name in actual_file_names
+
+    def test_get_dataset_identifier(self, schema_mapping_uri, dataset_mapping_uri):
+        schema_mapping = get_data_mapping(data_mapping_uri=schema_mapping_uri)
+        data_mapping = get_data_mapping(data_mapping_uri=dataset_mapping_uri)
+        json_schema = {
+                "$id": ("https://sage-bionetworks.github.io/mobile-client-json/"
+                        "schemas/v2/ArchiveMetadata.json")
+        }
+        schema_dataset_identifier = get_dataset_identifier(
```
thomasyu888, line 382: Assert the None use cases

tests/test_s3_to_json_s3.py
```diff
@@ -0,0 +1,477 @@
...
+                all_errors = validate_against_schema(
+                        data=j,
+                        schema=incorrect_metadata_json_schema,
+                        base_uri="https://sage-bionetworks.github.io/mobile-client-json/schemas/v2"
+                )
+                assert len(all_errors) == 1
+
+    def test_validate_data(self, s3_obj, archive_map, dataset_mapping_uri):
+        dataset_mapping = get_data_mapping(data_mapping_uri=dataset_mapping_uri)
+        validation_result = validate_data(
```
thomasyu888, line 432: NIt: Think about how once you've tested functions already, then you can mock the other functions.  That said, mocking requires maintenance.

tests/test_s3_to_json_s3.py
philerooski: I created a fixture for both `dataset_mapping.json` and `schema_mapping.json` that reads the mappings from a local path and does a direct comparison with the dict fetched here.

tests/test_s3_to_json_s3.py
philerooski: Fixed.

tests/test_s3_to_json_s3.py
philerooski: I split into different tests and renamed each test to be specific to its use case.

tests/test_s3_to_json_s3.py
philerooski: Done.

tests/test_s3_to_json_s3.py
philerooski: I'm using local fixtures to pass these mappings into the test functions now.

philerooski:
> Can you add a testing section in the readme to explain why you tested things with Docker instead of installing things into Github Actions?

I added a README with instructions on how to test. Setting up the dependencies (even locally!) is non-trivial so it's useful to have a Docker image to test both locally and as part of the CI/CD process.

tests/test_json_s3_to_parquet.py
```diff
@@ -0,0 +1,425 @@
...
+    def glue_database_name(self, namespace):
+        return f"{namespace}-pytest-database"
+
+    @pytest.fixture(scope="class")
+    def glue_nested_table_name(self):
+        return "dataset_pytest_nested_table"
+
+    @pytest.fixture(scope="class")
+    def glue_flat_table_name(self):
+        return "dataset_pytest_flat_table"
```
thomasyu888, line 25: Nit: These can technically be part of a `setup_class`  function.

```
def setup_class(self, namespace):
    self.glue_database_name = f"{namespace}-pytest-database"
    ....
```

You can then call these in the other functions without everything being fixtures.

tests/test_json_s3_to_parquet.py
```diff
@@ -0,0 +1,425 @@
...
+        glue_client.delete_crawler(Name=crawler_name)
+
+    @pytest.fixture()
+    def json_s3_objects(self, datadir, artifact_bucket, namespace):
+        s3_client = boto3.client("s3")
+        dataset_prefix = os.path.join(
+                "BridgeDownstream", namespace, "tests")
+        object_keys = []
+        for dirpath, _, filenames in os.walk(datadir):
+            if len(filenames) > 0:
```
thomasyu888, line 281: Cool trick:  you can just do

```
if filenames:
```
That does the same thing as the `len(filenames) > 0`

tests/test_json_s3_to_parquet.py
```diff
@@ -0,0 +1,425 @@
...
+                object_prefix = os.path.join(
+                        dataset_prefix,
+                        os.path.join(*dirpath.split("/")[-6:]))
```
thomasyu888, line 284: Nice use of the * here.

tests/test_json_s3_to_parquet.py
```diff
@@ -0,0 +1,425 @@
...
+            )
+        )
+        tables_with_index = {}
+        tables_with_index[glue_nested_table_name] = json_s3_to_parquet.add_index_to_table(
+                table_key=glue_nested_table_name,
+                table_name=glue_nested_table_name,
+                processed_tables=tables_with_index,
+                unprocessed_tables=nested_table_relationalized
+        )
+        assert (
```
thomasyu888, line 363: These seem to be separate test cases, but its ok they are in one giant test.

tests/README.md
```diff
@@ -0,0 +1,18 @@
+Tests are run inside a Docker container which includes all the necessary Glue/Spark dependencies and simulates the environment which the Glue jobs will be run in. A Dockerfile is included in this directory. To run tests locally, launch and attach to the docker container by referencing the following example command as a template:
```
thomasyu888, line 1: Might want to add that AWS credentials are required and that AWS resources will be created and torn down upon running this

tests/test_json_s3_to_parquet.py
philerooski: I prefer using fixtures.

tests/test_json_s3_to_parquet.py
philerooski: Nice. I'm going to leave it more explicit.

tests/README.md
philerooski: Done.

PR #95, opened 2022-10-12
[ETL-255] Explicit Glue resource permissions

philerooski:
Instead of using a managed policy, I've made the permissions that our Glue job and crawlers have explicit. I basically used the same permissions in the previously used managed policy, but I've modified some of them to better fit our pipeline (specifically, the S3 permissions). I also updated the Pipfile.lock, which should have been done in #94 .

### DIFF
--- a/templates/glue-job-role.yaml
+++ b/templates/glue-job-role.yaml
@@ -12,12 +12,46 @@ Parameters:
     Type: String
     Description: ARN of the SQS queue which we send invalid records to.
 
+  S3IntermediateBucketName:
+    Type: String
+    Description: Name of the S3 intermediate (JSON) bucket
+
+  S3ParquetBucketName:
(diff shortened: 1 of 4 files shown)


templates/glue-job-role.yaml
```diff
@@ -12,12 +12,46 @@ Parameters:
...
+        PolicyDocument:
+          Version: '2012-10-17'
+          Statement:
+          - Effect: Allow
+            Action:
+            - iam:ListRolePolicies
+            - iam:GetRole
+            - iam:GetRolePolicy
+            Resource:
+            - "*"
```
thomasyu888, line 54: Not a huge fan of the * but this was probably what the managed policies were doing.

thomasyu888 (approved):
Lgtm. Had one question

PR #96, opened 2022-10-13
Fix incorrect dependency in prod parquet bucket

philerooski:
There was an incorrect dependency specified in our prod stack for a very long time... it hasn't caused any issues until, apparently, now and is blocking #95 !

### DIFF
--- a/config/prod/s3-parquet-bucket.yaml
+++ b/config/prod/s3-parquet-bucket.yaml
@@ -1,7 +1,7 @@
 template_path: s3-bucket.yaml
 stack_name: bridge-downstream-parquet-bucket
 dependencies:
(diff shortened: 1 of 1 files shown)


thomasyu888 (approved):

thomasyu888:
Looks like you'll have to fix your tests though...

philerooski:
@thomasyu888 I found a few bugs that were causing inconsistent pass/fail behavior with the unit tests. I filed a separate JIRA [here](https://sagebionetworks.jira.com/browse/ETL-256) to address. Re-running the tests is enough to get them to pass right now, but addressing that JIRA should provide a fix going forward.

PR #99, opened 2022-10-21
Revert changes to config synapseBridgeDownstreamUserId parameter

philerooski:
Jinja and CFN do not play nicely here so we need to go back to explicitly specifying this parameter in the config file.


thomasyu888 (approved):
LGTM

PR #103, opened 2022-10-27
[ETL-256] Address unit test concurrency issues

philerooski:
Addresses:

* The glue resource files – specifically schema and dataset “legacy” mappings – should be uploaded to the artifact bucket either before the unit tests run or the tests should use mocked resources. (I've moved the upload step prior to the unit test step).

* The glue crawler should use the `json_s3_objects fixture` so that we know that the test data are in S3 before we run the crawler. (I made the crawler a standard, function-scoped fixture since we just need it to run once and have no need to reference it afterwards. This allows it to use the `json_s3_objects` fixture since it is no longer class-scoped.)


thomasyu888 (approved):
LGTM

PR #104, opened 2022-11-08
Add mtb-alpha stack to prod

philerooski:


thomasyu888 (approved):
LGTM.  I added this in the readme: https://sagebionetworks.jira.com/wiki/spaces/IBC/pages/2555281440/DataToParquet.  One question I had is where does the Synapse id  that Sonia provides us come into play

PR #105, opened 2022-11-15
[ETL-250] Update bootstrap trigger cron

philerooski:
Specified user on first line and added line break after last line (cron.d requirement)


thomasyu888 (approved):

thomasyu888:
@philerooski we probably need to revise the pre-commit to exclude the this file

philerooski:
@thomasyu888 Nice, I didn't know I could do that.

PR #112, opened 2023-01-03
[ETL-312] Ignore specific non-severe Android validation errors

philerooski:
* Ignores Android validation errors as outlined in ETL-312.
* Minor refactor of Github upload-and-deploy workflow.

[ETL-312]: https://sagebionetworks.jira.com/browse/ETL-312?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/src/glue/jobs/s3_to_json_s3.py
+++ b/src/glue/jobs/s3_to_json_s3.py
@@ -363,30 +363,31 @@ def is_expected_validation_error(validation_result, client_info):
     """
     if not validation_result["errors"]:
         return False
-    else:
-        if "Android" not in client_info:
+    if validation_result["appId"] != "mobile-toolbox":
+        return False
+    if "Android" not in client_info:
+        return False
+    if "metadata.json" in validation_result["errors"]:
+        metadata_errors = validation_result["errors"]["metadata.json"]
+        if not (len(metadata_errors) == 2 and
+                "'appName' is a required property" in metadata_errors and
+                "'files' is a required property" in metadata_errors):
             return False
-        if "metadata.json" in validation_result["errors"]:
-            metadata_errors = validation_result["errors"]["metadata.json"]
-            if not (len(metadata_errors) == 2 and
-                    "'appName' is a required property" in metadata_errors and
-                    "'files' is a required property" in metadata_errors):
-                return False
-        if "taskData.json" in validation_result["errors"]:
-            taskdata_errors = validation_result["errors"]["taskData.json"]
-            if not taskdata_errors == [
-                    "Additional properties are not allowed ('type' was unexpected)"]:
-                return False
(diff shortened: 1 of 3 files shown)


src/glue/jobs/s3_to_json_s3.py
```diff
@@ -363,30 +363,31 @@ def is_expected_validation_error(validation_result, client_info):
...
+        return False
+    if "Android" not in client_info:
+        return False
+    if "metadata.json" in validation_result["errors"]:
+        metadata_errors = validation_result["errors"]["metadata.json"]
+        if not (len(metadata_errors) == 2 and
+                "'appName' is a required property" in metadata_errors and
+                "'files' is a required property" in metadata_errors):
+            return False
+    if "taskData.json" in validation_result["errors"]:
```
thomasyu888, line 376: Nit. Use .get(....) then you don't need the extra if statement.

thomasyu888 (approved):
Lgtm - I will say that I do dislike the fact that we have many lines of conditions that return the same state (return false), BUT.... I can't really think of other ways to capture these conditions in a clean way that doesn't add some complexity to the code.

thomasyu888:
@rxu17 can you take a look at this too?

src/glue/jobs/s3_to_json_s3.py
philerooski: Normally, yes. But I'm checking for inequality (`if not errors == ...`) so unless I made the non-severe error the default return value of the `get` method then in the case that there were no taskData.json errors, the inequality would evaluate to true and `is_expected_validation_error` would prematurely return False. The way it's written now is more readable and less redundant imo than duplicating the non-severe errors in the arguments to `get`. Although, I think the `not ==` syntax is a bit unconventional for single comparisons, so I will change that.

src/glue/jobs/s3_to_json_s3.py
```diff
@@ -338,6 +338,58 @@ def validate_against_schema(data, schema):
...
+    if "weather.json" in validation_result["errors"]:
+        weather_errors = validation_result["errors"]["weather.json"]
+        if not weather_errors == [
+                "'type' is a required property"]:
+            return False
+    if "motion.json" in validation_result["errors"]:
+        motion_errors = validation_result["errors"]["motion.json"]
+        if not motion_errors == [
+                "'type' is a required property"]:
+            return False
```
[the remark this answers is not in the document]
philerooski: That's right. The complete list of errors is tracked here: https://sagebionetworks.jira.com/browse/BMC-432

I'll need to think about which of these qualifies as non-severe, but at least one of them which I've left out does.

PR #118, opened 2023-02-27
[ETL-355] Add prod stack for new study jfxqpk

rxu17:
Adds new study specific stack config and cron job for new synapse study jfxqpk


config/prod/studies/mobile-toolbox-jfxqpk.yaml
```diff
@@ -0,0 +1,30 @@
...
+  StudyName: jfxqpk
+  TemplateBucketName: {{ stack_group_config.artifact_bucket_name }}
+  ArtifactRef: {{ stack_group_config.latest_version }}
+  JsonBucketName: !stack_output_external bridge-downstream-intermediate-bucket::BucketName
+  ParquetBucketName: !stack_output_external bridge-downstream-parquet-bucket::BucketName
+  RoleArn: !stack_output_external glue-job-role::RoleArn
+  ClassifierName: !stack_output_external '{{ stack_group_config.namespace }}-array-of-records-classifier::ClassifierName'
+  SynapseAuthSsmParameterName: '{{ stack_group_config.synapseAuthSsmParameterName }}'
+  S3ToJsonS3JobName: !stack_output_external '{{ stack_group_config.namespace }}-glue-job-S3ToJsonS3::JobName'
+  JsonToParquetTriggerSchedule: 'cron(55 * * * ? *)'
```
philerooski, line 22: We already have a JSON to Parquet workflow with this schedule. I've run into issues in the past where if too many jobs are trying to write temp files or parquet datasets to the S3 bucket then we hit an API limit and the job fails. Probably not a big deal in this case, especially since it's overlapping with a QA study/project rather than an actual study, but let's bump this schedule by a few minutes.

philerooski (approved):
Minor change in comments

PR #119, opened 2023-02-27
[ETL-358] Remove expected validation errors in S3 to JSON S3 validation result

philerooski:
Removes anticipated (Android) validation errors from validation error result. This makes the validation process slightly more forgiving, but the only validation errors we ignore are concerned with missing required properties, superfluous properties, and a mis-specified value in motion ("acceleration" instead of "accelerometer" or "userAcceleration") -- so there shouldn't be any issues with mis-matched parquet dataset schemas.

Tested with the [data from the E2E QA study](https://www.synapse.org/#!Synapse:syn51121831/tables/) and [our test dataset](https://www.synapse.org/#!Synapse:syn29276125/datasets/).

### DIFF
--- a/src/glue/jobs/s3_to_json_s3.py
+++ b/src/glue/jobs/s3_to_json_s3.py
@@ -338,14 +338,14 @@ def validate_against_schema(data, schema):
     all_errors = [e.message for e in validator.iter_errors(data)]
     return all_errors
 
-def is_expected_validation_error(validation_result, client_info):
+def remove_expected_validation_errors(validation_result, client_info):
     """
     In the first year of MTB there were a number of issues with Android
     data not conforming to the JSON Schema. These aren't very severe
     inconsistincies (mostly missing or superfluous properties), but they
     prevent almost all Android data from being processed. This function
-    checks whether the validation errors fall into this non-severe category.
-    See ETL-312 for more details.
+    checks for and drops validation errors which fall into this non-severe
+    category. See ETL-312 and ETL-358 for more details.
 
     Args:
         validation_result (dict): A dictionary containing keys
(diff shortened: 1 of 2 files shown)


tests/test_s3_to_json_s3.py
```diff
@@ -599,9 +605,10 @@ def test_is_expected_validation_error(self):
                     "'stepPath' is a required property",
                 ]
         }
-        assert s3_to_json_s3.is_expected_validation_error(
+        unexpected_validation_errors = s3_to_json_s3.remove_expected_validation_errors(
                 validation_result=validation_result,
                 client_info=client_info)
+        assert unexpected_validation_errors == {}
```
rxu17, line 611: Should there be a test for an example with duplicated validation errors or is there code that drops the duplicates mentioned here: [ETL-358](https://sagebionetworks.jira.com/browse/ETL-358)?

rxu17 (approved):
LGTM! Just a small comment

src/glue/jobs/s3_to_json_s3.py
```diff
@@ -338,14 +338,14 @@ def validate_against_schema(data, schema):
     all_errors = [e.message for e in validator.iter_errors(data)]
     return all_errors
 
-def is_expected_validation_error(validation_result, client_info):
+def remove_expected_validation_errors(validation_result, client_info):
```
thomasyu888, line 341: Nit: has the return type been updated? Update the docstrings if so.

tests/test_s3_to_json_s3.py
philerooski: Good idea. Added.

PR #120, opened 2023-04-12
[ETL-417] Update crontab to use sharedschema_v1 parquet as reference

philerooski:
This affects the behavior of the bootstrap trigger script.

Rather than use the archivemetadata dataset as reference to determine which records have been processed, we will use the sharedschema dataset. This should result in fewer missing assessment data from the parquet datasets.

See [this comment](https://sagebionetworks.jira.com/browse/ETL-408?focusedCommentId=170177) for further context.


thomasyu888 (approved):
🚀  LGTM!

PR #121, opened 2023-04-14
Add helper script used to investigate ETL-408

philerooski:


src/scripts/JIRA/etl-408/etl_408.py
```diff
@@ -0,0 +1,172 @@
...
+        except Exception as e:
+            print("Encountered the following error while polling the SQS queue. "
+                  "Returning the messages collected thus far.\n")
+            print(str(e))
+            break
+    messages_df = pd.DataFrame(messages)
+    validation_result_df = pd.DataFrame.from_records(
+            messages_df.validation_result.values)
+    invalid_records = pd.concat([messages_df, validation_result_df], axis=1)
+    return invalid_records
```
philerooski, line 132: @thomasyu888 You can use this function to query an SQS queue.

src/scripts/JIRA/etl-408/etl_408.py
```diff
@@ -0,0 +1,172 @@
...
+    # This version from 2023-04-11
+    reference = pd.read_csv(syn.get(MISSING_DATA_REFERENCE).path)
+    study_reference = reference.query(
+            f"studyId == '{study}' and inStudyProject == 1 and inParquet == 0")
+    study_reference = study_reference.dropna(subset=["recordId"])
+    return study_reference
+
+
+def main():
+    aws_session = boto3.session.Session(region_name="us-east-1")
```
thomasyu888, line 154: @philerooski Thanks for putting this script together.  Is it correct to assume that to run this script, you would have to set up the login to the bridge prod account as the default AWS config?

src/scripts/JIRA/etl-408/etl_408.py
thomasyu888, line 1: Nit: I'm ok with this script since this is more of an exploration script as I imagine this can be refactored in the long run to used across the different BD studies along (maybe RECOVER?) and it should include:

* Adherence to pep8 standards
* argparsers to specify AWS config profile, and the globally defined constants
* Usage instructions

src/scripts/JIRA/etl-408/etl_408.py
philerooski: Yes, if you want to interact with the production account (or dev account, or any AWS account) you will need to configure your credentials.

src/scripts/JIRA/etl-408/etl_408.py
philerooski: I think there's a misunderstanding about the purpose of pushing code to `src/scripts/JIRA`. The intention is to preserve any one-off code we write that helped us with a specific JIRA ticket. The resulting script is not meant to be generalizable, production-ready, or even re-run.

Even so, I do lint these JIRA scripts before I submit a PR and every deviation from pep8 has been intentional.

thomasyu888 (approved):
👍🏼 LGTM!

PR #122, opened 2023-04-19
[ETL-425] Update bootstrap trigger to resolve missing data issue 

philerooski:
* Bootstrap trigger script now accepts a second parquet dataset and performs a union of the first and second dataset's "diff field" before taking a diff with the records on Synapse.
* Crontab updated to use both sharedschema and archivemetadata datasets.


philerooski:
Accidentally submitted the wrong branch.

PR #123, opened 2023-04-19
[ETL-425] Bootstrap trigger diffs upon union of sharedschema and archivemetadata

philerooski:
* Bootstrap trigger script now accepts a second parquet dataset and performs a union of the first and second dataset's "diff field" before taking a diff with the records on Synapse.
* Crontab updated to use both sharedschema and archivemetadata datasets.


src/scripts/bootstrap_trigger/bootstrap_trigger.py
```diff
@@ -226,8 +234,16 @@ def main():
                 dataset_uri=args.diff_s3_uri,
                 aws_session=aws_session,
                 columns=[args.diff_parquet_field])
+        already_processed_records = parquet_dataset[args.diff_parquet_field]
+        if args.diff_s3_uri_2 is not None and args.diff_parquet_field_2 is not None:
+            parquet_dataset_2 = get_parquet_dataset(
+                    dataset_uri=args.diff_s3_uri_2,
+                    aws_session=aws_session,
+                    columns=[args.diff_parquet_field_2])
+            already_processed_records = already_processed_records.append(
```
rxu17, line 243: Nit: Just realized this but there is a `FutureWarning: The series.append method is deprecated and will be removed from pandas in a future version. Use pandas.concat instead.` for the `append` method for pandas

```suggestion
            already_processed_records = pd.concat([already_processed_records, parquet_dataset_2[args.diff_parquet_field]])
```

rxu17 (approved):
LGTM, one minor comment!

src/scripts/bootstrap_trigger/bootstrap_trigger.py
philerooski: There was also a bug on that line so I fixed it.

rxu17: That extra `]` :')

philerooski: Oh, I meant in my code. I used the wrong command line argument.

rxu17: Wow, missed that as well. Nice catch :')

PR #124, opened 2023-04-24
[ETL-425 fix] Update bootstrap trigger to use correct set logic

philerooski:
If S is the set of records on Synapse, P1 the set of records in the first Parquet dataset and P2 the set of records in the second Parquet dataset, I was doing
```
S \ (P1 U P2)
```
When I should have been doing
```
(S \ P1) U (S \ P2)
```

### DIFF
--- a/src/scripts/bootstrap_trigger/bootstrap_trigger.py
+++ b/src/scripts/bootstrap_trigger/bootstrap_trigger.py
@@ -1,17 +1,14 @@
 """
 Submit Synapse files stored on an external S3 bucket to a Glue workflow.
-Of course, this assumes that the workflow has read permissions on the S3
-bucket.
 
-The --diff-* arguments allow a parquet dataset in a Synapse-linked S3 location to
-be diffed upon before submitting matching --query results to the --glue-workflow.
-In this way, data which has already been processed by the pipeline is not
-processed again.
+The --diff-* arguments allow one or two parquet dataset in a Synapse-linked S3 location to
+be diffed upon before submitting matching --query results from a Synapse table
+to the --glue-workflow. In this way, data which does not appear in the specified
+parquet datasets is submitted for processing.
 """
 import json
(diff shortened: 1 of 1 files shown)


src/scripts/bootstrap_trigger/bootstrap_trigger.py
```diff
@@ -235,20 +232,22 @@ def main():
                 dataset_uri=args.diff_s3_uri,
                 aws_session=aws_session,
                 columns=[args.diff_parquet_field])
-        already_processed_records = parquet_dataset[args.diff_parquet_field]
+        records_needing_processing = synapse_df.drop(
```
thomasyu888, line 235: Just so I'm understanding, the default isn't taking the union and you have to specify both diffs. 

1. Why don't we take the union by default?
2. Do you envision more diff fields to be added?

philerooski: > Why don't we take the union by default?

The union is taken "by default" if the user specifies two parquet datasets. Are you suggesting we remove the ability to only diff upon a single parquet dataset? What would it look like to take the union by default?

This is the current behavior:

* Specify no parquet datasets, and everything on Synapse is submitted
* Specify one parquet dataset, and the diff is taken upon that parquet dataset
* Specify two parquet datasets, and we submit the union of each parquet dataset diff'd upon Synapse.

> Do you envision more diff fields to be added?

Only the metadata and assessment data are required files, so it doesn't make sense to add others. You would submit for (re)processing all the records which didn't include a specific optional file (?) I can't think of a plausible use case.

thomasyu888: Ahh thanks for the explanation!  I misunderstood the default behavior. Approving

thomasyu888 (approved):
🔥 LGTM!

PR #126, opened 2023-04-27
Remove query flag from ccbcwq bootstrap trigger

philerooski:
There was no issue including the older records in processing.


thomasyu888 (commented):
Thanks just a small question prior to approving - can you help me understand by clarifying the description?  Are you saying:

* After you re-processed there were no issues and expected missing data were filled by the fix.

or


* This was an unnecessary change and wasn't the issue of missing data.

philerooski:
@thomasyu888 Both. But now that we have both Februrary and April data in the Parquet, if we don't also include the February data (as this PR does) then bootstrap trigger will break during the call to `pandas.DataFrame.drop`, since the February data will be missing from the index.

thomasyu888 (approved):
👍🏼 lgtm! That's unfortunate - seems like more debugging is in order then.

PR #128, opened 2023-05-11
[ETL-437] Add script to copy (underscore) S3 objects to new object

philerooski:
This script is meant to resolve the issues we've had processing JSON whose filename begins with an underscore. Originally, I wanted to move the objects by doing a copy followed by a delete, but according to [this doc](https://docs.aws.amazon.com/AmazonS3/latest/API/API_CopyObject.html) it might be complicated determining if a copy operation was completed successfully. So I decided to drop the "delete" action, this way if there are any issues we haven't inadvertently chucked the original data.

### DIFF
--- a/src/scripts/JIRA/etl-437/main.py
+++ b/src/scripts/JIRA/etl-437/main.py
@@ -1,0 +1,136 @@
+"""
+This script copies all objects under an S3 prefix which begin with
+a specific string to an object in the same "directory" but with
+the basename prefaced with another string.
+"""
+import json
+import os
+import argparse
+import boto3
+
+def read_args():
+    parser = argparse.ArgumentParser(
+            description="Copy S3 objects to another object prefaced with a string.")
+    parser.add_argument(
+            "--s3-bucket",
+            required=True,
+            help="The name of the S3 bucket."
+    )
+    parser.add_argument(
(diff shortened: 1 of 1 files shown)


src/scripts/JIRA/etl-437/main.py
```diff
@@ -0,0 +1,136 @@
...
+            help="Optional. The local AWS profile to use."
+    )
+    args = parser.parse_args()
+    return args
+
+def get_matching_objects(
+        s3_client: "botocore.client.S3",
+        bucket: str,
+        key_prefix: str,
+        match_string: str) -> "list[str]":
```
thomasyu888, line 45: Nit: I think for later versions of Python you can probably just do

```
-> list[str]
# but for older versions
from typing import List
-> List[str]
```

This is the same for dictionaries, but not a huge concern.

thomasyu888 (approved):
🔥 Going to pre-approve, but can you add in the description what command you ran?  What did you use as the `--preface-string`?

philerooski:
> can you add in the description what command you ran? What did you use as the --preface-string?

This script would be run multiple times, once for each study, so the arguments vary, but in the case of the `--preface-string` argument that's documented in the JIRA ticket.

PR #129, opened 2023-05-12
[ETL-440] Add script to validate self-referencing records on Synapse

philerooski:
This script validates record data against its schema by querying a Synapse file view for records. The records must use the self-referencing format in their metadata.json file. Any files in the record which do not have a `jsonSchema` property specified in their fileInfo object in metadata.json will not be validated. The validation results are written to `validation_errors.json`.

### DIFF
--- a/src/scripts/validate_synapse_records/validate_synapse_records.py
+++ b/src/scripts/validate_synapse_records/validate_synapse_records.py
@@ -1,0 +1,171 @@
+"""
+This script validates record data against its schema by querying
+a Synapse file view for records.
+
+The records must use the self-referencing format in their metadata.json file.
+Any files in the record which do not have a `jsonSchema` property specified in
+their fileInfo object in metadata.json will not be validated.
+
+The validation results are written to `validation_errors.json`.
+"""
+
+import argparse
+import json
+import zipfile
+import boto3
+import jsonschema
+import requests
+import synapseclient
+
+def read_args():
+    parser = argparse.ArgumentParser(
+            description="Validate records in a Synapse file view against their JSON Schema.")
+    parser.add_argument("--file-view",
+                        required=True,
+                        help=("The Synapse ID of a file view containing "
+                              "files to be submitted."))
+    parser.add_argument("--query",
+                        default="select * from {source_table}",
+                        help=("Optional. An f-string formatted query which filters the "
+                              "file view. Use {source_table} in the FROM clause. "
+                              "If this argument is not specified, all files in "
+                              "the --file-view will be submitted to the "
+                              "--glue-workflow."))
+    parser.add_argument("--ssm-parameter",
+                        help=("Optional. The name of the SSM parameter containing "
+                              "the Synapse personal access token. "
+                              "If not provided, cached credentials are used"))
+    parser.add_argument("--profile",
+                        help=("Optional. The AWS profile to use. Uses the default "
+                              "profile if not specified."))
+    args = parser.parse_args()
+    return args
+
+def get_synapse_client(ssm_parameter=None, aws_session=None):
+    """
+    Return an authenticated Synapse client.
+
+    Args:
+        ssm_parameter (str): Name of the SSM parameter containing the
+            BridgeDownstream Synapse password.
+        aws_session (boto3.session.Session)
+
+    Returns:
+        synapseclient.Synapse
+    """
+    if ssm_parameter is not None:
+        ssm_client = aws_session.client("ssm")
+        token = ssm_client.get_parameter(
+            Name=ssm_parameter,
+            WithDecryption=True)
+        syn = synapseclient.Synapse()
+        syn.login(authToken=token["Parameter"]["Value"])
+    else: # try cached credentials
+        syn = synapseclient.login()
+    return syn
+
+def get_schemas(synapse_files: list):
+    """
+    Get schemas for each JSON file in each Synapse file in a list of Synapse files.
+
+    Args:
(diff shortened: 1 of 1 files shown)


rxu17 (approved):
Looks good to me! I'm wondering why this is a one-time script and not something we'd forsee doing again?

philerooski:
@rxu17 Why do you say it's a one-time script?

philerooski:
That makes sense. I'll move it to `src/scripts/`

src/scripts/validate_synapse_records/validate_synapse_records.py
```diff
@@ -0,0 +1,172 @@
...
+import argparse
+import boto3
+import jsonschema
+import requests
+import synapseclient
+
+def read_args():
+    parser = argparse.ArgumentParser(
+            description="Validate records in a Synapse file view against their JSON Schema.")
+    parser.add_argument("--file-view",
```
thomasyu888, line 23: Just out of curiosity, since positional arguments are required, why not make this a positional argument?

src/scripts/validate_synapse_records/validate_synapse_records.py
```diff
@@ -0,0 +1,172 @@
...
+    Returns:
+        all_schemas (dict): A dictionary of schemas
+            {
+                "synapseId": {
+                    "filename": "schema_url",
+                    ...
+                },
+            }
+    """
+    all_schemas = {}
```
thomasyu888, line 83: Nit: look into collections.defaultdict.

thomasyu888 (approved):
🔥 LGTM!

going to pre-approve, just some comments. is your plan to incorporate this in the main pipeline (what's going to execute it?)

Also a test seems to be failing, but can we add tests for this script?

src/scripts/validate_synapse_records/validate_synapse_records.py
```diff
@@ -0,0 +1,172 @@
...
+        token = ssm_client.get_parameter(
+            Name=ssm_parameter,
+            WithDecryption=True)
+        syn = synapseclient.Synapse()
+        syn.login(authToken=token["Parameter"]["Value"])
+    else: # try cached credentials
+        syn = synapseclient.login()
+    return syn
+
+def get_schemas(synapse_files: list):
```
thomasyu888, line 67: Should we test this function?

src/scripts/validate_synapse_records/validate_synapse_records.py
```diff
@@ -0,0 +1,172 @@
...
+            if "metadata.json" in z.namelist():
+                with z.open("metadata.json", "r") as metadata_f:
+                    metadata = json.load(metadata_f)
+                    for file_info in metadata["files"]:
+                        if "jsonSchema" in file_info:
+                            all_schemas[synapse_file.id][file_info["filename"]] = \
+                                    file_info["jsonSchema"]
+    return all_schemas
+
+def validate_data(synapse_files: list, schemas: dict):
```
thomasyu888, line 96: Should we test this function?

src/scripts/validate_synapse_records/validate_synapse_records.py
philerooski: Why do you say that positional arguments are required? I find positional arguments to be relatively ambiguous and increase the likelihood of passing incorrect arguments, so I tend to always use flags.

src/scripts/validate_synapse_records/validate_synapse_records.py
philerooski: I dislike the idea of importing an additional library to cover such a simple use case.

philerooski:
@thomasyu888 I don't think it makes sense to write tests for this script. It's not used regularly – neither as part of study setup nor the pipeline itself. Tests create an additional maintenance burden for developers with little upside considering there's no need for the functionality provided by this script to be highly available. In other words, when we might want to use this script every 6 or so months, if something has broken in the meanwhile it's no issue to fix it on an as-needed basis.

src/scripts/validate_synapse_records/validate_synapse_records.py
thomasyu888: Hmmm, I misspoke, it seems like all arguments by default is None.   I agree with your usage of flags

thomasyu888:
@philerooski I am ok with your rationale.  We certainly have many things on DPE that we don't necessarily have tests for. What is your rationale behind this script being executed every 6 months?  Shouldn't this be executed on all new self referencing data?

TLDR: Tests are not meant to be convenient - it is an intentional maintenance burden, and tests also aren't only written if something is going to be made highly available.  If the code is modified "on accident" or "on purpose" the initial test written help provide the insight to how the function has been changed. If an error is found during the execution of this script, that is something you can write a test for so you make sure that particular error doesn't happen again.

philerooski:
> What is your rationale behind this script being executed every 6 months? Shouldn't this be executed on all new self referencing data?

This script actually covers a very narrow use case. Normally we can validate new assessments by updating the schema mapping and running new assessments through the pipeline. If we see the data in the JSON dataset then it's passed validation. In the case that's not possible (e.g., there's work that needs to be done on S3 to JSON before we can process this data, as was the case for schema refresh), we (or any assessment developer) can use this script to grab data directly from Synapse and validate the data.

So no, this won't be run on all new self-referencing data. It will be run at most every time we get a new schema from the assessment developers (~6 months perhaps). And it's only necessary to use this particular script if we find ourselves in the situation we can't run the data through the pipeline directly, as outlined in my previous paragraph.

PR #131, opened 2023-05-15
[ETL-426] Add new datasets as Glue tables and update crawlers

philerooski:


rxu17 (approved):
LGTM!

src/glue/resources/table_columns.yaml
```diff
@@ -1,4 +1,49 @@
 tables:
+  3DRotation_v1:
```
rxu17, line 2: Remind me again - how do we find the data types of the fields here for BD? Just through the Glue Crawler or parsing through the production 3D rotation data in Synapse ourselves (which would have all of the expected fields in one file or do you have to look through multiple)?

src/glue/resources/dataset_crawler_assignments.yaml
```diff
@@ -3,6 +3,10 @@ standard:
   - sharedSchema_v1
   - WeatherResult_v1
   - sharedSchema_for_Gradual_Onset_v1
+  - 3DRotation_v1
```
thomasyu888, line 6: Are we informed when new datasets are added by our collaborators or bridge?

src/glue/resources/table_columns.yaml
philerooski: Initially, we derive the data types from the JSON Schema provided for the assessment. These Glue table schemas can end up evolving beyond any single JSON Schema (for example, if the assessment developers put out a new revision of the 3DRotation assessment with an additional field and remove other fields), but that hasn't happened IRL yet.

src/glue/resources/dataset_crawler_assignments.yaml
philerooski: Short answer: yes. AFAIK nothing's been formalized (whatever that would look like), but everyone involved recognizes that BD needs to be part of the QA process. Sonia has been our link between the assessment developers (Northwestern, in this case) and ourselves so far.

thomasyu888 (approved):
🔥 LGTM!

PR #133, opened 2023-05-16
Fix syntax mistake in JSON to Parquet test

philerooski:


rxu17 (approved):
LGTM! Was this part of the reason for the [ETL-452](https://sagebionetworks.jira.com/browse/ETL-452) other than the deleting of the test glue resources?

philerooski:
No, this is unrelated. It worked fine as-is but the crawler name was ugly so I fixed it.

thomasyu888 (approved):

PR #134, opened 2023-05-17
[ETL-453] Add module to update Glue crawlers with additional targets

philerooski:
This module provides functions to update a glue crawler by adding additional s3 targets and (potentially) modifying the recrawl policy. If a glue crawler is configured to crawl new folders only, it cannot be updated via a regular stack update. Instead, we need to perform the following steps to completed the update:

1. Perform the desired crawler updates, but ensure that the update has a recrawl policy of crawl_everything so that the update completes successfully.
2. If the crawler originally had a different recrawl policy, we can "undo" the change to the recrawl policy we made in the previous step by doing an additional update -- but only to set the recrawl policy back to its original value

This script supports a specific use case which occurs when we add support for new datasets. Additional s3 targets must be added to the crawlers, but our crawlers recrawl policy is to only crawl new folders. So the default behavior of the function `add_target_to_crawler` is to add the s3 targets while overwriting the recrawl policy. We then update the crawler once more to change the recrawl policy back to crawl new folders only.

### DIFF
--- a/src/scripts/add_targets_to_crawler/add_targets_to_crawler.py
+++ b/src/scripts/add_targets_to_crawler/add_targets_to_crawler.py
@@ -21,6 +21,7 @@ We then update the crawler once more to change the recrawl policy back to crawl 
 folders only.
 """
 import argparse
+import logging
 import json
 import boto3
 
@@ -73,6 +74,11 @@ def read_args():
         default="default",
         help="Which AWS profile to use."
     )
+    parser.add_argument(
+        "--debug",
+        action="store_true",
+        help="Set the log level to debug."
+    )
     args = parser.parse_args()
     return args
 
@@ -115,7 +121,7 @@ def add_targets_to_crawler(
         crawler_name: str,
         s3_paths: list,
         record_state=False
-    ) -> dict:
+    ):
     """
     Adds a target to a Glue crawler.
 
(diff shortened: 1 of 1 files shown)


src/scripts/add_targets_to_crawler/add_targets_to_crawler.py
thomasyu888, line 1: Will this script be run automatically at some point?

philerooski:
@rxu17 and I had a discussion and decided on two changes:

1. Write the module to be invoked from the command line. I initially thought this would be too onerous passing a list of studies and S3 targets on the command line, but if we make some assumptions about how the S3 paths are formatted it's not too complicated.
2. Rather than assuming the crawlers have a recrawl policy of "crawl everything" or "crawl new folders only", we will cache the original recrawl policy when we get the crawler initially and reset the recrawl policy after adding the targets if it is not already "crawl everything".

src/scripts/add_targets_to_crawler/add_targets_to_crawler.py
philerooski: I initially thought that this would be too complicated, but it might be possible. Chat GPT tells me I can do something like:

```
name: Check file modification
        run: |
          # Replace 'path/to/file' with the actual path to your file
          if git diff --name-only HEAD^..HEAD | grep -q "path/to/file"; then
            echo "File has been modified"
          else
            echo "File has not been modified"
          fi
```

within the Glue workflow to check wether the `src/glue/resources/dataset_crawler_assignments.yaml` file has been modified. Do you think it's worth trying to do this work now? Or we can do a simpler stop gap solution by including running this script in the instructions for adding new datasets when we do our schema management work later this year. I'm not sure if the time spent incorporating this into the CI/CD is a priority right now.

philerooski:
@thomasyu888 @rxu17 I've updated the module to be used as a command line tool. Going one step further and incorporating this tool into the workflow would add another ~50% of time spent working on this. The main difficulty I see is being able to programmatically generate a list of study identifiers which we process data for. We can:

- Authenticate with Synapse and query https://www.synapse.org/#!Synapse:syn50615998
- Parse the study identifiers out of `config/prod/studies/`.

But both of these solutions only work for the production environment and neither is something I've done from a Github workflow before, so the simplicity of the action may belie the actual complexity of implementation. Develop we would hardcode to only update `namespace=bridge-downstream`.

src/scripts/add_targets_to_crawler/add_targets_to_crawler.py
```diff
@@ -0,0 +1,200 @@
...
+    )
+    parser.add_argument(
+        "--studies",
+        nargs="+",
+        help="List of study identifiers (e.g., cxhnxd)"
+    )
+    parser.add_argument(
+        "--datasets",
+        nargs="+",
+        help="List of dataset identifiers (e.g., ArchiveMetadata_v1)"
```
thomasyu888, line 53: I think you alluded a little to this, but what would it take to also validate that the studies and datasets are specified correctly?

thomasyu888 (approved):
👍🏼 LGTM! Thanks @philerooski , we can create a ticket to discuss/incorporate this into the workflow in the future.  I'm happy with this implementation as long as it's documented in an SOP somewhere to run this code.  Going to pre-approve, but just a question.

rxu17 (approved):
LGTM!

PR #136, opened 2023-05-23
[ETL-458] Fixes for S3 to JSON, Glue tables

philerooski:
This fixes the mistake I introduced when overwriting the json schema `$id`. I made some other fixes to make the job less dumb. Details in the comments.

I ran the new assessments and the test dataset through this pipeline and read all the parquet datasets produced from the JSON, so I'm pretty confident this works.

### DIFF
--- a/src/glue/jobs/s3_to_json_s3.py
+++ b/src/glue/jobs/s3_to_json_s3.py
@@ -10,6 +10,7 @@ derived from the $id property of the associated JSON Schema (if the file
 has a JSON schema) or mapped to directly from the assessment ID/revision
 and filename for certain older assessments.
 """
+import copy
 import io
 import json
 import logging
@@ -128,7 +129,9 @@ def get_archive_map(archive_map_version):
 
 def update_json_schemas(s3_obj, archive_map, json_schemas):
     """
-    Get JSON Schemas for all files in a zipped S3 object.
+    Get JSON Schemas for all files in a zipped S3 object (if the schema exists).
+    If a file does not have a JSON Schema, it will not be included in the
+    return object.
 
     Args:
         s3_obj (dict): An S3 object as returned by boto3.get_object.
@@ -150,6 +153,7 @@ def update_json_schemas(s3_obj, archive_map, json_schemas):
             file_metadata = {
                 "file_name": file_name,
                 "app_id": app_id,
+                "record_id": s3_obj["Metadata"]["recordid"],
                 "assessment_id": assessment_id,
                 "assessment_revision": assessment_revision,
             }
@@ -159,13 +163,14 @@ def update_json_schemas(s3_obj, archive_map, json_schemas):
                 json_schemas=json_schemas,
                 self_ref_schema_list=self_ref_schema_list,
             )
-            novel_schema = True
-            for cached_schema in json_schemas:
-                if all([cached_schema[k] == json_schema[k] for k in file_metadata]):
-                    novel_schema = False
-                    break
-            if novel_schema:
-                json_schemas.append(json_schema)
+            if json_schema["schema"] is not None:
+                novel_schema = True
+                for cached_schema in json_schemas:
+                    if cached_schema["schema"]["$id"] == json_schema["schema"]["$id"]:
+                        novel_schema = False
+                        break
+                if novel_schema:
+                    json_schemas.append(json_schema)
     return json_schemas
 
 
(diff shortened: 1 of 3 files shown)


src/glue/jobs/s3_to_json_s3.py
```diff
@@ -159,13 +161,14 @@ def update_json_schemas(s3_obj, archive_map, json_schemas):
...
-            if novel_schema:
-                json_schemas.append(json_schema)
+            if json_schema["schema"] is not None:
+                novel_schema = True
+                for cached_schema in json_schemas:
+                    if cached_schema["schema"]["$id"] == json_schema["schema"]["$id"]:
+                        novel_schema = False
+                        break
+                if novel_schema:
+                    json_schemas.append(json_schema)
```
philerooski, line 171: Rather than match schemas using the keys in `file_metadata` (which doesn't actually make much sense... not sure why I did it this way in the first place), I compare the schemas directly by their `$id`.

src/glue/jobs/s3_to_json_s3.py
```diff
@@ -196,8 +200,8 @@ def get_json_schema(archive_map, file_metadata, json_schemas, self_ref_schema_li
             * url (str)
             * schema (dict)
             * app_id (str)
-            * assessment_id (int)
-            * assessment_revision (int)
+            * assessment_id (str)
+            * assessment_revision (str)
```
philerooski, line 204: These type changes were both corrections and make `file_metadata` more consistent across the functions.

src/glue/jobs/s3_to_json_s3.py
```diff
@@ -397,8 +400,11 @@ def validate_against_schema(data, schema):
     # This is a workaround for this bug
     # https://github.com/python-jsonschema/jsonschema/issues/1012
     if "$id" in schema and schema["$id"].startswith("schemas/v0/"):
-        schema["$id"] = ""
-    validator = validator_cls(schema=schema)
+        substitute_schema = copy.deepcopy(schema)
+        substitute_schema["$id"] = ""
+        validator = validator_cls(schema=substitute_schema)
+    else:
+        validator = validator_cls(schema=schema)
```
philerooski, line 407: This is the bug fix. This function referenced a pointer to the schema in `json_schemas`, so when I overwrote `$id` I was overwriting the actual schema in `json_schemas`, and not just a local copy. So when it came time to look up the dataset identifier, `$id` was blank. I explicitly make a copy here so that we don't have this problem.

src/glue/resources/table_columns.yaml
```diff
@@ -201,9 +201,9 @@ tables:
       - Name: startDate
         Type: string
       - Name: stepHistory
-        Type: struct<endDate:string,identifier:string,interactions:struct<controlEvent:string,parent:string,stepIdentifier:string,timestamp:string,userInteractionIdentifier:string,userInteractionNewStyleEncoder:boolean,value:string>,position:integer,practice:boolean,response:integer,responseTime:double,score:integer,startDate:string,type:string,wasInterrupted:boolean>
+        Type: struct<endDate:string,identifier:string,interactions:struct<controlEvent:string,parent:string,stepIdentifier:string,timestamp:string,userInteractionIdentifier:string,userInteractionNewStyleEncoder:boolean,value:string>,position:int,practice:boolean,response:int,responseTime:double,score:int,startDate:string,type:string,wasInterrupted:boolean>
```
philerooski, line 204: This should be `int`, not `integer`

src/glue/jobs/s3_to_json_s3.py
philerooski: Ah, so before this change I was even adding files without schemas to `json_schemas` – which still doesn't make sense, but it explains why I was comparing the keys in `file_metadata`.

tests/test_s3_to_json_s3.py
```diff
@@ -445,12 +445,8 @@ def test_update_json_schemas(self, s3_obj, archive_map):
         )
         assert isinstance(json_schemas, list)
         file_names = [
-            "taskData",
             "taskData.json",
-            "info.json",
             "motion.json",
-            "weather.json",
-            "taskResult.json",
```
philerooski, line 453: These files are in the mocked `s3_obj`, but not in the mocked `archive_map`. Now that we are no longer including files without JSON schemas in `json_schemas`, we don't expect these files.

src/glue/jobs/s3_to_json_s3.py
thomasyu888: I might have asked before, but why are you writing the schema as a blank string?

src/glue/resources/table_columns.yaml
thomasyu888: Good catch, is that because "integer" isn't supported?

src/glue/jobs/s3_to_json_s3.py
```diff
@@ -150,6 +151,7 @@ def update_json_schemas(s3_obj, archive_map, json_schemas):
             file_metadata = {
                 "file_name": file_name,
                 "app_id": app_id,
+                "record_id": s3_obj["Metadata"]["recordid"],
```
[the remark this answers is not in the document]
thomasyu888: Can you speak to how it makes it more consistent?

thomasyu888 (approved):
🚀 LGTM!  Great bug fixes - I'm going to pre-approve, but please address the comments I made.

src/glue/resources/table_columns.yaml
philerooski: Glue tables want integer types to be written as `int`.

tests/test_s3_to_json_s3.py
philerooski: Yes, but not for the `update_json_schemas` function. If a file doesn't have a JSON Schema, we either ignore it or map it using `dataset_mapping.yaml` (if we can find it there). We test that functionality in `test_get_dataset_identifier_does_not_have_schema`.

src/glue/jobs/s3_to_json_s3.py
philerooski: I updated the docstring to describe this behavior.

PR #140, opened 2023-05-24
[ETL-314] Add parameter for deployment environment to manage artifacts script

philerooski:
* I added a `--environment` flag for specifying the deployment environment for the `manage_artifacts` script. This makes it a lot easier to deploy to either dev or prod. I also did quite a bit of cleaning up of the `manage_artifacts` script because it was pretty rough around the edges.
* I incorporated this new parameterization into our upload-and-deploy workflow. Believe it or not, artifacts like the schema mapping and glue job scripts were not being automatically uploaded as part of our deployment until now.

### DIFF
--- a/.github/workflows/upload-and-deploy.yaml
+++ b/.github/workflows/upload-and-deploy.yaml
@@ -45,6 +45,10 @@ jobs:
       - name: Copy files to templates bucket
         run: python src/scripts/manage_artifacts/artifacts.py --upload --ref $NAMESPACE
 
+        # Hardcode --ref as stopgap for actual job script versioning
+      - name: Copy Glue scripts to templates bucket
+        run: python src/scripts/manage_artifacts/artifacts.py --upload --ref "v0.1"
+
   pytest-docker:
     name: Build and push testing docker image to ECR
     needs: upload-files
@@ -201,6 +205,13 @@ jobs:
           aws_access_key_id: ${{ secrets.CI_USER_ACCESS_KEY_ID }}
           aws_secret_access_key: ${{ secrets.CI_USER_SECRET_ACCESS_KEY }}
           role_to_assume: ${{ secrets.CI_ROLE_TO_ASSUME }}
+
+      - name: Copy files to templates bucket
+        run: python src/scripts/manage_artifacts/artifacts.py --environment $environment --upload --ref $NAMESPACE
+
+        # Hardcode --ref as stopgap for actual job script versioning
+      - name: Copy Glue scripts to templates bucket
+        run: python src/scripts/manage_artifacts/artifacts.py --environment $environment --upload --ref "v0.1"
 
       - name: Login to Amazon ECR
         id: login-ecr
--- a/src/scripts/manage_artifacts/artifacts.py
+++ b/src/scripts/manage_artifacts/artifacts.py
@@ -1,78 +1,86 @@
+"""
+Manage cloudformation artifacts for a specific ref (namespace)
+"""
 import argparse
 import subprocess
 
-cfn_bucket = 'sceptre-cloudformation-bucket-bucket-65ci2qog5w6l'
-repo_name = 'BridgeDownstream'
+REPO_NAME = 'BridgeDownstream'
 
 
 def read_args():
-  descriptions = '''
-  Uploading to S3 and deletion from S3 of scripts and templates.
-  '''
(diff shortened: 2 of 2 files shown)


.github/workflows/upload-and-deploy.yaml
```diff
@@ -202,6 +206,13 @@ jobs:
           aws_secret_access_key: ${{ secrets.CI_USER_SECRET_ACCESS_KEY }}
           role_to_assume: ${{ secrets.CI_ROLE_TO_ASSUME }}
 
+      - name: Copy files to templates bucket
```
rxu17, line 209: I believe you mentioned that in BD we share the `v0.1` namespace. Is this section just setting us up for the future when we use the namespaced scripts?

src/scripts/manage_artifacts/artifacts.py
```diff
@@ -1,78 +1,86 @@
...
+  s3_path = f's3://{cfn_bucket}/{REPO_NAME}/'
   cmd = ['aws','s3','ls', s3_path]
   execute_command(cmd)
 
 
-def main(args):
+def main():
+  args = read_args()
+  if args.environment == "prod":
+    cfn_bucket = "sceptre-cloudformation-bucket-bucket-10mwvvuhlvtk9"
```
rxu17, line 75: Out of curiosity, was there a significance to the prod/develop cfn buckets named this way?

rxu17 (approved):
LGTM! Just a couple of questions

.github/workflows/upload-and-deploy.yaml
philerooski: No, this just propagates the existing design of having the Glue job scripts put in the v0.1 namespace and everything else being put in the "actual" namespace.

src/scripts/manage_artifacts/artifacts.py
philerooski: No, these buckets were created by IT along with the account. They were already there so we started using them for some of our artifact stuff.

.github/workflows/upload-and-deploy.yaml
```diff
@@ -45,6 +45,10 @@ jobs:
       - name: Copy files to templates bucket
         run: python src/scripts/manage_artifacts/artifacts.py --upload --ref $NAMESPACE
 
+        # Hardcode --ref as stopgap for actual job script versioning
+      - name: Copy Glue scripts to templates bucket
+        run: python src/scripts/manage_artifacts/artifacts.py --upload --ref "v0.1"
```
thomasyu888, line 50: Why is it v0.1 here?  Is this in reference to github release versions?

src/scripts/manage_artifacts/artifacts.py
```diff
@@ -1,78 +1,86 @@
...
-  descriptions = '''
-  Uploading to S3 and deletion from S3 of scripts and templates.
-  '''
   parser = argparse.ArgumentParser(
-    description='')
-  parser.add_argument(
-    '--ref')
+    description='Uploading to S3 and deletion from S3 of scripts and templates.',
+    formatter_class=argparse.ArgumentDefaultsHelpFormatter)
+  parser.add_argument('--ref')
```
thomasyu888, line 14: Nit: It might be helpful to add some `help` docs to these so that we can look back and know what each of these parameters are.

.github/workflows/upload-and-deploy.yaml
philerooski: Back when Tess and I were first working on Glue jobs we began using this crude versioning system so that different versions of the S3 to JSON job could potentially be deployed simultaneously. The idea was that older studies would need older implementations whereas newer studies could use the latest and greatest implementation. There hasn't been a real need yet for supporting multiple versions of the S3 to JSON job, but the versioning system has stuck around. The glue jobs have their job scripts store to this "v0.1" namespace, which is separate from where we store other artifacts like schema mapping.

philerooski: Eventually, I'd like to re-design this.

src/scripts/manage_artifacts/artifacts.py
philerooski: I added some additional documentation.

thomasyu888 (approved):
🔥 LGTM!

PR #141, opened 2023-05-25
Clean up "matrix" component of upload-and-deploy workflow

philerooski:
Fix for broken action resulting from pulling in #140


thomasyu888 (approved):
🔥 Nice simplification!

PR #144, opened 2023-05-30
[ETL-463] Add stack for study gxvwhj

philerooski:


thomasyu888 (changes requested):
Do you need to add in the cronjob?

thomasyu888 (approved):
🔥 nice! Going to pre-approve, but just to confirm, from now on, we will not be using shared_metadataschema right?

philerooski:
@thomasyu888 I think you mean `sharedSchema`. Only these ICAR assessments use the self-referencing format so far. So for studies which only use ICAR assessments they won't have a `sharedSchema` dataset. But all other assessments thus far use `sharedSchema` and studies which include them will have a `sharedSchema` dataset.

thomasyu888:
Ah yes @philerooski, that is what I meant. Thanks!

PR #146, opened 2023-06-29
[ETL-316] Map Android's microphone.json to a dataset identifier

philerooski:
This change works by having any `microphone.json` files pose as `microphone_levels.json` files in the S3 to JSON job. Hence, for the purpose of getting the JSON schema (and corresponding dataset identifier) and validating against a JSON schema, a `microphone.json` file is treated as a `microphone_levels.json` file.

The one adverse effect of this change is that sometimes the file will be reported as `microphone_levels.json` in logging statements, but this isn't likely to cause too much confusion since a record will contain `microphone.json` xor `microphone_levels.json`. It was not strictly necessary to update the `dataset_mapping.json` file (since the "posing" happens before checking this file), but it does make explicit to any future developers that `microphone.json` is mapped to `AudioLevelRecord_v1`.


src/glue/jobs/s3_to_json_s3.py
rxu17, line 1: Should there be any additions to the tests for the microphone.json -> microphone_level.json?

src/glue/jobs/s3_to_json_s3.py
```diff
@@ -150,6 +150,8 @@ def update_json_schemas(s3_obj, archive_map, json_schemas):
         self_ref_schema_list = get_self_ref_schema_list(z)
         for json_path in contents:
             file_name = os.path.basename(json_path)
+            if file_name == "microphone.json":
```
rxu17, line 153: Could there be a comment somewhere briefly explaining this change?

src/glue/jobs/s3_to_json_s3.py
rxu17, line 1: Will Android data "correct" this incorrect file name in the future?

rxu17 (approved):
LGTM! Just a few comments

thomasyu888 (approved):
🔥 LGTM!  Pre-approving but please address Rixing's comments.

src/glue/jobs/s3_to_json_s3.py
philerooski: Whoops, I keep forgetting to update the unit tests.

src/glue/jobs/s3_to_json_s3.py
philerooski: I think a more effective method of tracking the reasons behind changes is with git blame. Anyone can see that this change is associated with this PR, and can track the Jira ticket from there.

src/glue/jobs/s3_to_json_s3.py
philerooski: They already have, although the background recorders on Android have been disabled ever since we discovered that Android data with background recorders often had upload issues. This fix is for retroactive data.

src/glue/jobs/s3_to_json_s3.py
```diff
@@ -374,7 +378,10 @@ def validate_data(s3_obj, archive_map, json_schemas, dataset_mapping):
             with z.open(json_path, "r") as p:
                 logger.debug(
                         "Validating %s from %s against %s",
-                        file_name, validation_result["recordId"], json_schema["url"])
```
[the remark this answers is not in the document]
philerooski: The logging library recommends the use of lazy-string formatting, and I think most linters (including the one I'm using: pylint) will produce a warning if f-strings are used. https://docs.python.org/3/howto/logging.html#optimization

In this case, it's not exactly mission critical to use lazy formatting, but that's my reasoning.

src/glue/jobs/s3_to_json_s3.py
rxu17: Yes I like that, I remember seeing that before.

src/glue/jobs/s3_to_json_s3.py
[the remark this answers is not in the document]
thomasyu888: Thanks Phil! I learn something new every day - agree with your rationale.

PR #147, opened 2023-07-12
Update S3 to JSON job to use v4.6.1 of archive-map.json

philerooski:


thomasyu888 (approved):
💯 should it be part of a process somewhere to create a ticket for us to update this whenever the version is bumped?

philerooski:
@thomasyu888 It's unlikely that archive-map.json will be updated in the future considering that we are moving towards self-referencing data.

PR #150, opened 2023-08-25
[ETL-485] Check if test resource exists before creating

philerooski:
If it does, delete the resource (or detach and delete in the case of an IAM role).

### DIFF
--- a/tests/test_json_s3_to_parquet.py
+++ b/tests/test_json_s3_to_parquet.py
@@ -38,13 +38,19 @@ class TestJsonS3ToParquet:
     @pytest.fixture(scope="class", autouse=True)
     def glue_database(self, glue_database_name, glue_database_path):
         glue_client = boto3.client("glue")
-        glue_database = glue_client.create_database(
-                DatabaseInput={
-                    "Name": glue_database_name,
-                    "Description": "A database for pytest unit tests.",
-                    "LocationUri": glue_database_path
-                }
-        )
+        potential_entity_conflict = True
+        while potential_entity_conflict:
+            try:
+                glue_database = glue_client.create_database(
+                        DatabaseInput={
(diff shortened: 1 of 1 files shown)


thomasyu888 (approved):
💯 nice - why have we not run into this before?

philerooski:
@thomasyu888 We have, but we've manually deleted the resources before rerunning tests in the past.

tests/test_json_s3_to_parquet.py
```diff
@@ -38,13 +38,19 @@ def glue_database_path(self, artifact_bucket, namespace):
...
         glue_client = boto3.client("glue")
-        glue_database = glue_client.create_database(
-                DatabaseInput={
-                    "Name": glue_database_name,
-                    "Description": "A database for pytest unit tests.",
-                    "LocationUri": glue_database_path
-                }
-        )
+        potential_entity_conflict = True
+        while potential_entity_conflict:
```
rxu17, line 42: Do we want a timeout condition for the while loops here?

rxu17 (approved):
LGTM! just a comment

tests/test_json_s3_to_parquet.py
philerooski: I think the only way we could get stuck in a doom loop is if the entity already exists, we try to delete it, but deleting the entity fails for some reason. If it fails we will either get an exception (not an `AlreadyExistsException`, but something else) - in which case the program will terminate - or the method will fail silently. If it continues to fail silently we will get stuck deleting (and silently failing) over and over again.

But these functions will never fail silently because they read the HTTP code from the server and if something is off they will raise an exception. If the server never returns a response then... either these methods have their own timeout or we would never reach the part of our code with our own timeout condition.

PR #151, opened 2023-08-25
[ETL-481] Utilize glue's resolve choice in dynamic frame input in JSON to Parquet

rxu17:
**Purpose:** This PR casts the fields in a dynamic frame object input to the data types in the schema crawled by the glue crawler in our glue tables. This solves the issue of glue splitting a field out into multiple columns thinking it has more than one data type.

This PR has three new core functions that convert from the hive schema format to json schema format, and then parses it to get the data types from the glue tables crawled by the glue crawlers so we can use it in the glue function `resolveChoice`.
- parse_hive_schema
- convert_json_schema_to_specs
- cast_glue_data_types (calls the other two)

A couple of caveats are that:

- We only expect to convert fields in a non-array format so something like: `fieldName: [1, 2, 3]` won't be resolved 
- We also won't be able to resolve choice in a field to be an array or a struct

You can see these conditions handled in the `convert_json_schema_to_specs` functions

This is a draft. I still need to add a timeout condition for the while loops/recursive functionality. One question is if the placement of the **cast_glue_data_types** function is in the expected spot in the main json to parquet code.


src/glue/jobs/json_s3_to_parquet.py
```diff
@@ -132,105 +141,246 @@ def add_index_to_table(table_key, table_name, processed_tables, unprocessed_tabl
...
                 succinct_name = c.replace(field_prefix, "").replace(".", "_")
                 # If key is a duplicate we keep the original field name
                 if succinct_name in df_with_index.columns:
                     continue
-                df_with_index = df_with_index.withColumnRenamed(
-                        c, succinct_name)
+                df_with_index = df_with_index.withColumnRenamed(c, succinct_name)
     return df_with_index
 
+
```
thomasyu888, line 180: Nice one!

src/glue/jobs/json_s3_to_parquet.py
```diff
@@ -132,105 +141,246 @@ def add_index_to_table(table_key, table_name, processed_tables, unprocessed_tabl
...
+    specs = []
+    for field_info in glue_table_schema:
+        json_schema = parse_hive_schema(
+            hive_str=field_info["Type"], top_level_field=field_info["Name"]
+        )
+        specs.extend(convert_json_schema_to_specs(json_schema))
+    glue_table = dynamic_frame.resolveChoice(specs=specs)
+    return glue_table
+
+
```
thomasyu888, line 214: Nit: is there a python package that does this already?

src/glue/jobs/json_s3_to_parquet.py
```diff
@@ -132,105 +141,246 @@ def add_index_to_table(table_key, table_name, processed_tables, unprocessed_tabl
...
+    to_parse = hive_str
+    parents = []
+    curr_elem = root
+
+    key = None
+
+    iteration = 0
+    while to_parse:
+        if iteration > MAX_ITERATIONS:
```
thomasyu888, line 246: How did you determine the max iterations?

src/glue/jobs/json_s3_to_parquet.py
```diff
@@ -132,105 +141,246 @@ def add_index_to_table(table_key, table_name, processed_tables, unprocessed_tabl
...
+            raise Exception(
+                f"Parsing hive schema reached maximum iterations of {MAX_ITERATIONS}"
+            )
+        left, operator, to_parse = r.match(to_parse).groups()
+
+        if operator == "struct<" or operator == "array<":
```
thomasyu888, line 252: Nit: `If key is not None` to be precise.

src/glue/jobs/json_s3_to_parquet.py
```diff
@@ -132,105 +141,246 @@ def add_index_to_table(table_key, table_name, processed_tables, unprocessed_tabl
...
+        dict: python dictionary format of input hive struct
+    """
+    r = re.compile(r"(.*?)(struct<|array<|[:,>])(.*)")
+    root = dict()
+
+    # adjust hive string to have top level field in the hive str
+    hive_str = f"struct<{top_level_field}:{hive_str}>"
+
+    to_parse = hive_str
+    parents = []
```
thomasyu888, line 239: Hmm, I may be missing something, but is this line required?

src/glue/jobs/json_s3_to_parquet.py
```diff
@@ -132,105 +141,246 @@ def add_index_to_table(table_key, table_name, processed_tables, unprocessed_tabl
...
+    while to_parse:
+        if iteration > MAX_ITERATIONS:
+            raise Exception(
+                f"Parsing hive schema reached maximum iterations of {MAX_ITERATIONS}"
+            )
+        left, operator, to_parse = r.match(to_parse).groups()
+
+        if operator == "struct<" or operator == "array<":
+            parents.append(curr_elem)
+            new_elem = dict() if operator == "struct<" else list()
```
thomasyu888, line 254: I'm a little confused here, you set the key and then you write over the existing dict with an empty one?

src/glue/jobs/json_s3_to_parquet.py
```diff
@@ -132,105 +141,246 @@ def add_index_to_table(table_key, table_name, processed_tables, unprocessed_tabl
...
+        path (list, optional): Path used in recursion in function.
+            Defaults to [].
+
+    Returns:
+        list: list of tuples in the expected form:
+         [
+             (path_to_field1 : cast:<field1_type>),
+             (path_to_field2 : cast:<field2_type>)
+         ]
+    """
```
thomasyu888, line 292: Will we only need to support json with length one?

src/glue/jobs/json_s3_to_parquet.py
```diff
@@ -132,105 +141,246 @@ def add_index_to_table(table_key, table_name, processed_tables, unprocessed_tabl
...
+            if left:
+                if isinstance(curr_elem, dict):
+                    curr_elem[key] = left
+                elif isinstance(curr_elem, list):
+                    curr_elem.append(left)
+
+            if operator == ">":
+                curr_elem = parents.pop()
+        iteration += 1
+    return root
```
thomasyu888, line 274: Its a bit risky to have an empty list as a default value: https://www.codiga.io/blog/python-avoid-empty-array-parameter/

thomasyu888 (commented):
Nice work! Leaving some initial comments.

tests/Dockerfile
```diff
@@ -1,4 +1,4 @@
 FROM amazon/aws-glue-libs:glue_libs_3.0.0_image_01
 
-RUN pip3 install synapseclient==2.5.* pytest-datadir
+RUN pip3 install synapseclient==2.5.* pyspark==3.* pyarrow==9.* pytest-datadir
```
thomasyu888, line 3: Nit: may want to use the ~ syntax

src/glue/jobs/json_s3_to_parquet.py
rxu17: I just set it to 100 for now because I don't imagine we'll have a super nested schema to parse? I could also keep it consistent and just set it to the same as the default recursion limit that's already in python: `sys.getrecursionlimit()` which is 1000.

src/glue/jobs/json_s3_to_parquet.py
rxu17: Are you referring to `parse_hive_schema`? I did look for one and as a result of my research it sounds like there wasn't anything that does this (converting from json hive schema string to dictionary format) out of the box and it has to be manually done in this way. I tried looking into some of the custom parsing libraries like `pyparsing` but they didn't seem to support this use case.

src/glue/jobs/json_s3_to_parquet.py
rxu17: JSON with length one? Nope this is in theory supposed to support a json schema that has many nested fields

src/glue/jobs/json_s3_to_parquet.py
```diff
@@ -21,19 +22,22 @@
 from awsglue.utils import getResolvedOptions
 from pyspark import SparkContext
 
+
+# maximum iterations for any loop in order to timeout a loop
+MAX_ITERATIONS = 100
```
philerooski, line 27: Why is this defined globally? This seems like it should be a default parameter for the `parse_hive_schema` function.

src/glue/jobs/json_s3_to_parquet.py
```diff
@@ -132,105 +141,246 @@ def add_index_to_table(table_key, table_name, processed_tables, unprocessed_tabl
...
+                elif isinstance(curr_elem, list):
+                    curr_elem.append(left)
+
+            if operator == ">":
+                curr_elem = parents.pop()
+        iteration += 1
+    return root
+
+
+def convert_json_schema_to_specs(json_schema: dict, path: list = []) -> list:
```
philerooski, line 277: We already have the concept of a "JSON schema" in other components of BD, but the `json_schema` you are referring to here is a very different thing. Let's rename this (`dict_schema`?) so that we don't get our wires crossed in the future.

You refer to `json_schema` in other parts of this script so be sure to update those references as well.

src/glue/jobs/json_s3_to_parquet.py
```diff
@@ -132,105 +141,246 @@ def add_index_to_table(table_key, table_name, processed_tables, unprocessed_tabl
...
+
+        if operator == "struct<" or operator == "array<":
+            parents.append(curr_elem)
+            new_elem = dict() if operator == "struct<" else list()
+            if key:
+                curr_elem[key] = new_elem
+                curr_elem = new_elem
+            elif isinstance(curr_elem, list):
+                curr_elem.append(new_elem)
+                curr_elem = new_elem
```
philerooski, line 260: What's happening in these two conditions? It looks like you either add to a dict or list, but then you proceed to overwrite that dict or list with `new_elem`?

src/glue/jobs/json_s3_to_parquet.py
```diff
@@ -132,105 +141,246 @@ def add_index_to_table(table_key, table_name, processed_tables, unprocessed_tabl
...
+        glue_context=glue_context,
     )
+
+    # cast table's glue types to crawlers' json schema types
+    if args["cast_glue_data_types"]:
+        table = cast_glue_data_types(
+            dynamic_frame=table,
+            glue_database_name=table_name,
+            glue_table_name=workflow_run_properties["database"],
+        )
```
philerooski, line 345: You should be replacing L81-83 with the `cast_glue_data_types` logic. Otherwise we've already resolved the types (incorrectly) in `get_table` above.

tests/README.md
```diff
@@ -11,6 +11,15 @@ Note that the docker image included in the above example command is publicly acc
...
 
+If you are trying to run glue tests particularly the ones in the test_json_s3_to_parquet.py be sure to configure your aws credentials prior to running the below:
+
+```
+aws configure set aws_access_key_id $AWS_ACCESS_KEY_ID
+aws configure set aws_secret_access_key $AWS_SECRET_ACCESS_KEY
+aws configure set aws_session_token $AWS_SESSION_TOKEN
+aws configure set region $AWS_REGION
+```
+
```
philerooski, line 22: You can also have the credentials configured as your default profile in the credentials file and mount your credentials file:

```
docker run ...
-v ~/.aws:/home/glue_user/.aws
```
probably it's enough to say "configure your AWS credentials" and let developers read AWS documentation if they don't know what that means.

src/glue/jobs/json_s3_to_parquet.py
philerooski: Ohh, I think I understand what's happening. You're updating the pointer. Is there a way to make this clearer in the code?

philerooski (changes requested):

src/glue/jobs/json_s3_to_parquet.py
rxu17: Ah that was what I wasn't sure about it in the end, because I thought the `get_table`'s resolveChoice wouldn't permanently cast it

src/glue/jobs/json_s3_to_parquet.py
rxu17: Which line? `curr_elem = root`? Not sure why, but with my github view of the comments, I'm not sure which line(s) the comments are pointing at

